PR #1, opened 2023-06-21
ci: Add linters

fischeti:


util/lint/waiver.verible
colluca, line 1: @fischeti I think it would be better to convert these to [in-file waiver comments](https://github.com/chipsalliance/verible/tree/master/verilog/tools/lint#in-file-waiver-comments). This would avoid that the hardcoded lines in the waiver file break as soon as the files are edited.

colluca (changes requested):

util/lint/waiver.verible
fischeti: Done, I changed all waivers with specific lines to in-line waivers

colluca (approved):

PR #4, opened 2023-07-04
ci: Fix Makefile target dependencies and extend simutils.py

GiannaP:
Contains various general fixes:
- simutils.py: Capture missing binary error and print also successful tests
- gitlab-ci: Fix path to run.py for tests
- update `target/snitch_cluster/Makefile` `clean` targets
- bender: Move `VSIM_SOURCES`, `VLT_SOURCES` depending on bender target from `common.mk` to `target/snitch_cluster/Makefile`
- gitlab-ci: Fix VCS target dependency such that the required files inside the `generated` directory are actually generated before the compilation
- sw: add runtime compilation for Banshee to github CI

### DIFF
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -24,7 +24,6 @@ TB_SRCS   := $(wildcard ${ROOT}/hw/ip/test/*.sv)
 TB_DIR    := ${ROOT}/target/common/test
 
 VSIM_BENDER   += -t test -t rtl -t simulation -t vsim
-VSIM_SOURCES   = $(shell ${BENDER} script flist ${VSIM_BENDER} | ${SED_SRCS})
 VSIM_BUILDDIR := work-vsim
 
 # fesvr is being installed here
@@ -44,7 +43,6 @@ VLT_FLAGS    += -Wno-UNOPTFLAT
 VLT_FLAGS    += -Wno-fatal
 VLT_FLAGS    += --unroll-count 1024
 VLT_BENDER   += -t rtl
-VLT_SOURCES  := $(shell ${BENDER} script flist ${VLT_BENDER} | ${SED_SRCS})
 VLT_CFLAGS   += -std=c++14 -pthread
 VLT_CFLAGS   +=-I ${VLT_BUILDDIR} -I $(VLT_ROOT)/include -I $(VLT_ROOT)/include/vltstd -I $(VLT_FESVR)/include -I $(TB_DIR)
 
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -26,6 +26,11 @@ TB_DIR    := ${ROOT}/target/common/test
 VSIM_BENDER   += -t test -t rtl -t simulation -t vsim
 VSIM_BUILDDIR := work-vsim
 
+# VCS_BUILDDIR should to be the same as the `DEFAULT : ./work-vcs`
+# in target/snitch_cluster/synopsys_sim.setup
+VCS_BENDER   += -t test -t rtl -t simulation -t vcs
+VCS_BUILDDIR := work-vcs
+
 # fesvr is being installed here
 FESVR          ?= ${MKFILE_DIR}work
 FESVR_VERSION  ?= 35d50bc40e59ea1d5566fbd3d9226023821b1bb6
@@ -159,11 +164,11 @@ endef
 #######
 # VCS #
 #######
-work-vcs/compile.sh: ${VSIM_SOURCES} ${TB_SRCS}
-	mkdir -p work-vcs
-	${BENDER} script vcs ${VSIM_BENDER} --vlog-arg="${VLOGAN_FLAGS}" --vcom-arg="${VHDLAN_FLAGS}" > $@
+$(VCS_BUILDDIR)/compile.sh:
+	mkdir -p $(VCS_BUILDDIR)
+	${BENDER} script vcs ${VCS_BENDER} --vlog-arg="${VLOGAN_FLAGS}" --vcom-arg="${VHDLAN_FLAGS}" > $@
 	chmod +x $@
-	$@ > work-vcs/compile.log
+	$@ > $(VCS_BUILDDIR)/compile.log
 
 ########
 # Util #
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -24,17 +24,21 @@ TB_SRCS   := $(wildcard ${ROOT}/hw/ip/test/*.sv)
 TB_DIR    := ${ROOT}/target/common/test
 
 VSIM_BENDER   += -t test -t rtl -t simulation -t vsim
+VSIM_SOURCES   = $(shell ${BENDER} script flist ${VSIM_BENDER} | ${SED_SRCS})
 VSIM_BUILDDIR := work-vsim
 
 # VCS_BUILDDIR should to be the same as the `DEFAULT : ./work-vcs`
 # in target/snitch_cluster/synopsys_sim.setup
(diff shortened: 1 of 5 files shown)


target/snitch_cluster/Makefile
```diff
@@ -223,7 +231,7 @@ bin/snitch_cluster.vlt: $(VLT_AR) $(VLT_COBJ) ${VLT_BUILDDIR}/lib/libfesvr.a
 
 # Clean all build directories and temporary files for Questasim simulation
 clean-vsim: clean-work
-	rm -rf bin/snitch_cluster.vsim bin/snitch_cluster.vsim.gui work-vsim vsim.wlf
+	rm -rf bin/snitch_cluster.vsim bin/snitch_cluster.vsim.gui $(VSIM_BUILDDIR) vsim.wlf
 
 ${VSIM_BUILDDIR}/compile.vsim.tcl: $(VSIM_SOURCES) ${TB_SRCS} ${TB_CC_SOURCES}
```
colluca, line 236: Similarly to what you did for VCS, I would move the `$(VSIM_SOURCES)` and `$(TB_SRCS)` as dependencies of the `bin/snitch_cluster.vsim` target instead.

target/common/common.mk
```diff
@@ -24,9 +24,13 @@ TB_SRCS   := $(wildcard ${ROOT}/hw/ip/test/*.sv)
 TB_DIR    := ${ROOT}/target/common/test
 
 VSIM_BENDER   += -t test -t rtl -t simulation -t vsim
-VSIM_SOURCES   = $(shell ${BENDER} script flist ${VSIM_BENDER} | ${SED_SRCS})
```
colluca, line 27: I would keep the `XXX_SOURCES` variables here, since they are used the same also in Occamy. Is there a conflicting reason to move them to the target-specific Makefile?

target/snitch_cluster/Makefile
```diff
@@ -243,12 +251,12 @@ bin/snitch_cluster.vsim: ${VSIM_BUILDDIR}/compile.vsim.tcl work/lib/libfesvr.a
...
-	rm -rf bin/snitch_cluster.vcs work-vcs
+	rm -rf bin/snitch_cluster.vcs $(VCS_BUILDDIR) vc_hdrs.h
 
 # Build compilation script and compile all sources for VCS simulation
-bin/snitch_cluster.vcs: work-vcs/compile.sh work/lib/libfesvr.a $(GENERATED_DIR)/snitch_cluster_wrapper.sv ${TB_DIR}/common_lib.cc $(GENERATED_DIR)/bootdata.cc
+bin/snitch_cluster.vcs: ${VCS_SOURCES} ${TB_SRCS} $(VCS_BUILDDIR)/compile.sh work/lib/libfesvr.a $(GENERATED_DIR)/snitch_cluster_wrapper.sv ${TB_DIR}/common_lib.cc $(GENERATED_DIR)/bootdata.cc
 	mkdir -p bin
-	vcs -Mlib=work-vcs -Mdir=work-vcs -o bin/snitch_cluster.vcs -cc $(CC) -cpp $(CXX) \
+	vcs -Mlib=$(VCS_BUILDDIR) -Mdir=$(VCS_BUILDDIR) -o bin/snitch_cluster.vcs -cc $(CC) -cpp $(CXX) \
 		-assert disable_cover -override_timescale=1ns/1ps -full64 tb_bin ${TB_DIR}/rtl_lib.cc ${TB_DIR}/common_lib.cc $(GENERATED_DIR)/bootdata.cc \
```
colluca, line 260: We can replace `${TB_DIR}/rtl_lib.cc ${TB_DIR}/common_lib.cc $(GENERATED_DIR)/bootdata.cc` with `$(TB_CC_SRCS)`.

target/snitch_cluster/Makefile
```diff
@@ -243,12 +251,12 @@ bin/snitch_cluster.vsim: ${VSIM_BUILDDIR}/compile.vsim.tcl work/lib/libfesvr.a
...
 		-CFLAGS "-std=c++14 -I${MKFILE_DIR} -I${MKFILE_DIR}/test -I${FESVR}/include -I${TB_DIR}" -LDFLAGS "-L${FESVR}/lib" -lfesvr
```
colluca, line 261: We can replace `${TB_DIR}/rtl_lib.cc ${TB_DIR}/common_lib.cc $(GENERATED_DIR)/bootdata.cc` with `$(TB_CC_SRCS)`, and `"-std=c++14 -I${MKFILE_DIR} -I${MKFILE_DIR}/test -I${FESVR}/include -I${TB_DIR}"` with `$(TB_CC_FLAGS)`.

target/common/common.mk
colluca, line 27: I would keep the `XXX_SOURCES` variables here, since they are used the same also in Occamy. Is there a conflicting reason to move them to the target-specific Makefile?

colluca (changes requested):

target/common/common.mk
GiannaP: yes, it was needed because the `Makefile` inside the `snitch_cluster` target sets `VSIM_BENDER += -t snitch_cluster`. This additional target flag will not be taken for the bender execution here, and hence the snitch_cluster specific files will be missing from the list.

util/simutils.py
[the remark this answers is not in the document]
GiannaP: You mean the `print_failed_test` and `print_success_test` functions that are called inside the `print_test_summary`?

I actually really like the categorization of the failing tests. It gives you also when you run the tests locally a direct feedback on what went wrong. E.g. SW compilation failed, or HW was not compiled successfully or similar. I think we do not need to cover all cases, but such general failures unrelated to the functional correctness are nice to cover IMO.
But I agree, for the successful tests I would remove the return code printing tbh.

target/common/common.mk
GiannaP: I also like the feedback on which tests passed. This gives you, e.g., when running tests locally, a direct feedback that the tests you wanted to run actually were executed. But as mentioned above, I would not print the individual return codes for the passing tests.

colluca: That shouldn't actually be an issue for the `VSIM_SOURCES` as defined previously.
https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html

I do agree that this is a problem with the `VLT_SOURCES` if we ever decide to extend it in the target directory. This can be corrected by defining it as a recursively expanded variable instead of a simply extended variable.

It seems to work from my tests, I can push this change.

util/simutils.py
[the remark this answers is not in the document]
colluca: I mean recording all the information on the test failure to later report it in the test summary.

I think it's not really useful as one would anyways debug on test at a time, and scrolling up in the log you can find much more detailed and useful information than in the summary.

So I would keep the summary just to know where you have to scroll back, i.e. which log to inspect further, so the script can be simplified dramatically.

target/common/common.mk
GiannaP: Ah nice, I was not aware of that feature. Yes, sure go ahead with it :)

util/simutils.py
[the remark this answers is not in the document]
GiannaP: Well, I often used these tests locally also while developing occamy back in the day. If you run it locally to check your changes you have the log directly in the shell which makes scrolling really annoying. I really liked this summary feature for local tests in the snitch repo and will really miss it tbh :)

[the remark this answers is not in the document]
colluca: But actually should be really easy also locally: you can just search in the shell for the test name. I added that at the end of the test as well, so you don't have to scroll down all the vsim output

PR #10, opened 2023-07-11
printf: Fix printf support for new runtime

GiannaP:
Add lost RTL runtime `putchar` implementation and add `simple_printf` test.

**Attention**: The test does currently not check the actual printed results after the simulation!!!


colluca (approved):
LGTM :)

paulsc96 (approved):

PR #26, opened 2023-07-13
snDNN: Move kernels into library compatible with new Runtime

paulsc96:
Reopen of #5.


viv-eth (changes requested):
`torch` module is missing the requirements.txt. Can you add it? I think then tests will pass

GiannaP:
> `torch` module is missing the requirements.txt. Can you add it? I think then tests will pass

I had to re-build the docker on the specific branch such that the newest snDNN requirements are installed on Docker.

GiannaP:
After the re-base on the newest master, the simulation of `fusedconv` reports errors. Therefore, I have removed it from the list called `sw/apps/passing-apps.list` that is executed in the CI.

PR #27, opened 2023-07-13
hw: Improve debug parameterization, fix debug CSR RW

paulsc96:
This includes the contributions from #7 and further gates debug features inside the Snitch core with parameters, controlled by the existing `debug_enable` configuration key:

* It adds software RW to the `dcsr`, `dpc`, and `dscratch0` registers.
* It disables the instantiation of any debug registers, and thus logic through constant propagation, when `debug_enable` is 0.


colluca (approved):

colluca:
Should we perhaps add an assertion which raises an error or a warning if we get a `irq_i.debug` when `DebugSupport` is 0?

PR #29, opened 2023-07-14
tutorial: Correct layout file

colluca:
Ten cores was legacy from single-cluster Occamy. Here we have only nine cores.

### DIFF
--- a/target/snitch_cluster/README.md
+++ b/target/snitch_cluster/README.md
@@ -244,8 +244,8 @@ The traces contain a lot of information which we might not be interested at firs
 
 ```
             , compute
-"range(0,9)",       1
-9           ,
+"range(0,8)",       1
+8           ,
 
 ```
 
@@ -255,9 +255,9 @@ Then run the following commands:
 # Similar to logs/perf.csv but filters all but tstart and tend metrics
 make logs/event.csv
 # Labels, filters and reorders the event regions as specified by an application-specific layout file
-../../../util/trace/layout_events.py logs/event.csv sw/apps/axpy/layout.csv -o logs/trace.csv
+../../util/trace/layout_events.py logs/event.csv sw/apps/axpy/layout.csv -o logs/trace.csv
 # Creates a trace file which can be visualized with Chrome's TraceViewer
-../../../util/trace/eventvis.py -o logs/trace.json logs/trace.csv
+../../util/trace/eventvis.py -o logs/trace.json logs/trace.csv
 ```
 
 Go to `http://ui.perfetto.dev/`. Here you can load the `logs/trace.json` file and graphically view the runtime of the compute region in your code. To learn more about the layout file syntax and what the Python scripts do you can have a look at the description comment at the start of the scripts themselves.


target/snitch_cluster/README.md
```diff
@@ -244,8 +244,8 @@ The traces contain a lot of information which we might not be interested at firs
 
 ```
             , compute
-"range(0,9)",       1
-9           ,
+"range(0,8)",       1
```
GiannaP, line 247: I am only just learning about the feature and format. But given the example, would it not make more sense to make it to illustrate the feature?

```
            , compute, idle   
"range(0,7)",       1,
8           ,        ,       1
```

target/snitch_cluster/README.md
```diff
@@ -244,8 +244,8 @@ The traces contain a lot of information which we might not be interested at firs
...
+8           ,
 
 ```
```
GiannaP, line 251: The following two lines need to remove one hierarchy `../`:

```
# Labels, filters and reorders the event regions as specified by an application-specific layout file
../../util/trace/layout_events.py logs/event.csv sw/apps/axpy/layout.csv -o logs/trace.csv
# Creates a trace file which can be visualized with Chrome's TraceViewer
../../util/trace/eventvis.py -o logs/trace.json logs/trace.csv
```

GiannaP (changes requested):

target/snitch_cluster/README.md
colluca: I think this would indeed make sense. Thanks for the suggestion @GiannaP :)

colluca: After looking at the code however, the idle region would actually be very short (it doesn't include the barrier synchronization with the other cores).

To not complicate the code at this point I would postpone this demonstration. I anyways wanted to extend the tutorial to demonstrate how to use the DMA, and that would be perfect to also demonstrate the feature you suggested here.

GiannaP: Yes, exactly. It looks like that because there is no barrier with the other cores:
image

Nevertheless, it gives a first functionality indication, and the DMA core is not computing anything in that phase, so it is still better fitting than calling it also a computation phase for the DMA core IMO. However, feel free to postpone this for future documentation.

target/snitch_cluster/README.md
GiannaP: It is 2-3 lines below the `layout.csv` example in the same documentation file `target/snitch_cluster/README.md`. As it is not in the "modified code" area, I can not mark it exactly, but when testing it I noticed the instructions only need two `../../util....` instead of three `../../../util.....`. Thought it might make sense to fix it in this PR aswell :)

target/snitch_cluster/README.md
colluca: Indeed I agree, but in fact the currently modified `layout.csv` should not display anything for the DM core, i.e. neither "compute" nor "idle".

GiannaP: Ah, in your new version, yes, indeed, it does not. I also just noticed Thread 7 is missing in my example. As mentioned, feel free to leave it for now.

target/snitch_cluster/README.md
colluca: Good catch! Thanks :)

I made the suggested corrections.

GiannaP (approved):
LGTM :)

PR #30, opened 2023-07-20
Added 3 polybench benchmarks to snitch: Covariance, Correlation and Atax

jpcf:
Do not merge it just now. I will continue working on this branch a bit longer. Just created the PR to set authorship.

### DIFF
--- a/sw/polybench/.gitignore
+++ b/sw/polybench/.gitignore
@@ -1,0 +1,3 @@
+debug*
+build/
+data/
--- a/sw/polybench/Makefile
+++ b/sw/polybench/Makefile
@@ -1,5 +1,5 @@
 APP = polybench
-SRCS = src/covariance.c
+SRCS = src/correlation.c
 INCDIRS = data
 
 include ../common.mk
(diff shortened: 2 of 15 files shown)


colluca:
> Do not merge it just now. I will continue working on this branch a bit longer. Just created the PR to set authorship.

Great. Then please convert it to a draft PR until then :)

paulsc96 (commented):
Sorry for redrafting this; did not see it was undrafted.

Do I understand correctly that this uses an Occamy cluster configuration with FDIV/FSQRT enabled? I am currently exploring some architectural choices, but maybe we should converge on a single standard config for software.

colluca:
> Do I understand correctly that this uses an Occamy cluster configuration with FDIV/FSQRT enabled? I am currently exploring some architectural choices, but maybe we should converge on a single standard config for software.

That is correct. But it was only convenient because with the LLVM toolchain we cannot disable lowering to FDIV and FSQRT instructions (cfr. GCC's `-mno-fdiv` flag). However, with Federico's recent extension we can revert to the default hardware configuration. Let's discuss this further in the next compiler meeting.

colluca:
Succeeded by https://github.com/pulp-platform/snitch_cluster/pull/111.

PR #31, opened 2023-07-28
banshee: Fix Banshee runtime support and add tests

GiannaP:
- fix Banshee runtime
- add small documentation on dependency tools (banshee, bender, spike)
- run all existing tests with Banshee and check the passing and failing ones
- add tests to Github CI
- add tests to Gitlab CI
- Fix VCS and VLT simulation setup

TODO:
- [x] `banshee` repo: Make `flexfloat` submodule based on https instead of ssh for simplified docker recursive repository checkout


target/snitch_cluster/README.md
```diff
@@ -45,18 +48,26 @@ make CFG_OVERRIDE=cfg/custom.hjson bin/snitch_cluster.vlt
...
+
+# for Banshee simulation (requires slightly different runtime)
+make SELECT_RUNTIME=banshee DEBUG=ON sw
 ```
 
 The `sw` target first generates some C header files which depend on the hardware configuration. Hence, the need to generate the software for the same configuration as your hardware. Afterwards, it recursively invokes the `make` target in the `sw` subdirectory to build the apps/kernels which have been developed in that directory.
 
 The `DEBUG=ON` flag is used to tell the compiler to produce debugging symbols. It is necessary for the `annotate` target, showcased in the Debugging section of this guide, to work.
 
+The `SELECT_RUNTIME` is set by default to `rtl`. To use the Banshee runtime, it requires to be set to be set to `banshee` with `SELECT_RUNTIME=banshee`.
```
paulsc96, line 69: This sentence repeats "to be set". Also, it's a bit hard to follow.

paulsc96 (approved):
Overall, LGTM

.github/workflows/ci.yml
```diff
@@ -79,4 +96,28 @@ jobs:
...
+      - name: Build Software
+        run: |
           make -C target/snitch_cluster SELECT_RUNTIME=banshee sw
+      - name: Run Unit Tests
+        working-directory: target/snitch_cluster
+        env:
+          SNITCH_LOG: info
+        run: |-
+          ./sw/tests/run.py sw/tests/banshee-apps.list --simulator banshee \
+          2>&1 | tee tests.log
```
colluca, line 108: Why do we need to tee the output to a file and grep the file to verify that the tests passed?

.gitlab-ci.yml
```diff
@@ -14,6 +14,10 @@ variables:
   QUESTA: questa-2022.3
   LLVM_BINROOT: /usr/pack/riscv-1.0-kgf/pulp-llvm-0.12.0/bin
   CLANG: /usr/pack/riscv-1.0-kgf/pulp-llvm-0.12.0/bin/clang
+  CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER: /usr/pack/gcc-9.2.0-af/linux-x64/bin/gcc
+  LLVM_SYS_120_PREFIX: /usr/pack/llvm-12.0.1-af
+  SNITCH_LOG: info
+  CMAKE: cmake-3.18.1
```
colluca, line 20: Do we still need CMake?

.gitlab-ci.yml
```diff
@@ -88,24 +111,77 @@ snitch-cluster-vlt:
   script:
     - cd target/snitch_cluster
     - $VERILATOR make bin/snitch_cluster.vlt
-    - $VERILATOR ./sw/test/run.py sw/tests/passing-apps.list --simulator verilator
-    - $VERILATOR ./sw/apps/run.py sw/apps/passing-apps.list --simulator verilator
+    - $VERILATOR ./sw/tests/run.py sw/tests/passing-apps.list --simulator verilator 2>&1 | tee tests.log
```
colluca, line 114: See question above

.gitlab-ci.yml
```diff
@@ -88,24 +111,77 @@ snitch-cluster-vlt:
...
-    - $QUESTA ./sw/apps/run.py sw/apps/passing-apps.list --simulator vsim
+    - $QUESTA ./sw/tests/run.py sw/tests/passing-apps.list --simulator vsim 2>&1 | tee tests.log
+    - $QUESTA ./sw/apps/run.py sw/apps/passing-apps.list --simulator vsim 2>&1 | tee apps.log
+    - grep -zoP 'All tests passed' tests.log
+    - grep -zoP 'All tests passed' apps.log
+  # yamllint enable rule:line-length
+  artifacts:
+    when: on_failure
+    paths:
+      - target/snitch_cluster/tests.log
```
colluca, line 160: This output is on the CI console. Why would we also want to artifact it?

sw/snRuntime/src/start.c
```diff
@@ -84,12 +84,6 @@ static inline void snrt_init_cls() {
 
 static inline void snrt_init_libs() { snrt_alloc_init(); }
 
-static inline void snrt_exit(int exit_code) {
-    extern volatile uint32_t tohost;
-
-    if (snrt_global_core_idx() == 0) tohost = (exit_code << 1) | 1;
-}
-
```
colluca, line 92: This has to stay, otherwise the call to it below will give an error at compile time.

.github/workflows/ci.yml
GiannaP: Because the python return value was not forwarded to the CI runner. E.g. also on Gitlab VCS never really run and all tests failed, but it was reported as successful. This makes it a very explicit check that cannot be broken without being noticed.

.gitlab-ci.yml
GiannaP: The flexfloat submodule of Banshee needs it, yes.

target/snitch_cluster/Makefile
```diff
@@ -253,8 +253,8 @@ clean-vcs: clean-work
 bin/snitch_cluster.vcs: ${VCS_SOURCES} ${TB_SRCS} $(TB_CC_SOURCES) $(VCS_BUILDDIR)/compile.sh work/lib/libfesvr.a
 	mkdir -p bin
 	vcs -Mlib=$(VCS_BUILDDIR) -Mdir=$(VCS_BUILDDIR) -o bin/snitch_cluster.vcs -cc $(CC) -cpp $(CXX) \
-		-assert disable_cover -override_timescale=1ns/1ps -full64 tb_bin $(TB_CC_SRCS) \
-		-CFLAGS $(TB_CC_FLAGS) -LDFLAGS "-L${FESVR}/lib" -lfesvr
+		-assert disable_cover -override_timescale=1ns/1ps -full64 tb_bin $(TB_CC_SOURCES) \
+		-CFLAGS "$(TB_CC_FLAGS)" -LDFLAGS "-L${FESVR}/lib" -lfesvr
```
colluca, line 257: How is it that this worked before?

.gitlab-ci.yml
GiannaP: Good point, I think I added it before I piped the output to the log file and the CI shell simultaneously. Let's remove it.

target/snitch_cluster/README.md
```diff
@@ -31,6 +31,9 @@ vcs-2020.12 make bin/snitch_cluster.vcs
 
 These commands compile the RTL sources respectively in `work-vlt`, `work-vsim` and `work-vcs`. Additionally, common C++ testbench sources (e.g. the [frontend server (fesvr)](https://github.com/riscv-software-src/riscv-isa-sim)) are compiled under `work`. Each command will also generate a script or an executable (e.g. `bin/snitch_cluster.vsim`) which you can invoke to simulate the hardware. We will see how to do this in a later section.
 
+### Building the Banshee simulator
+Instead of running an RTL simulation, you can use our instruction-accuracte simulator called `banshee`. To install the simulator, please follow the instructions of the Banshee repository: [https://github.com/pulp-platform/banshee](https://github.com/pulp-platform/banshee).
```
colluca, line 35: Typo: accurate

docs/ug/getting_started.md
```diff
@@ -74,3 +74,15 @@ Install the required packages in the currently active virtual environment:
...
+
+This repository has a couple of tool dependencies:
+
+- Dependency management tool `bender`:
+   ```bash
+   # Bender tool
+   cargo install bender --version 0.23.2
+   ```
+- Instruction-accurate simulator `banshee`:
+  Follow instructions on [Banshee Github Repo](https://github.com/pulp-platform/banshee)
```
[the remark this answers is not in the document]
GiannaP: I am aware of groups that use snitch outside of the docker for their own research where I had to help with the dependencies etc. So ideally, the documentation also covers the installation of every single tool that is needed. So IMO this should be added and extended in the future as it reduces our personal support for partners.

target/snitch_cluster/README.md
colluca: Would it be possible to have Banshee installed on IIS machines? If not, I would add "For Docker container users, Banshee is already installed.", otherwise we can remove this altogether

target/snitch_cluster/README.md
```diff
@@ -45,18 +48,26 @@ make CFG_OVERRIDE=cfg/custom.hjson bin/snitch_cluster.vlt
 
 ___Note:__ whenever you override the configuration file on the `make` command-line, the configuration will be stored in the `cfg/lru.hjson` file. Successive invocations of `make` will automatically pick up the `cfg/lru.hjson` file. You can therefore omit the `CFG_OVERRIDE` definition in successive commands unless you want to override the least-recently used configuration._
 
+Banshee uses also a cluster configuration file, however, that is given directly when simulating a specific binary with banshee with the help of the `--configuration <cluster_config.yaml>`.
```
colluca, line 51: "with the help of the `--configuration <cluster_config.yaml>` option" or just "with the help of `--configuration <cluster_config.yaml>`"

target/snitch_cluster/README.md
colluca: "The `SELECT_RUNTIME` flag" or just "`SELECT_RUNTIME`"

sw/snRuntime/src/start.c
GiannaP: It still compiles (see CI). I moved it to [target/snitch_cluster/sw/runtime/rtl/src/snitch_cluster_start.c](https://github.com/pulp-platform/snitch_cluster/pull/31/files#diff-2989eedc7e2459373210b51f82afebfb9c02a0623ac4ccce51191b7aa9803aae). In that way it can be overwritten in the Banshee runtime with Banshee-specific return. Previously the two exit functions for RTL and Banshee had different names which prevented the OMP runtime (which explicitly calls `snrt_exit` in it's runtime part) from running on Banshee.

target/snitch_cluster/Makefile
GiannaP: It actually did not -> the CI failed but it was not watched because the Python return value is not forwarded to CI. See comment at the beginning.

target/snitch_cluster/sw/runtime/banshee/src/snitch_cluster_start.c
colluca, line 1: To be reviewed in light of the comment to "start.c"

target/snitch_cluster/sw/runtime/banshee/src/snrt.h
```diff
@@ -25,9 +26,12 @@
 #include "dm.h"
 #include "dma.h"
 #include "eu.h"
+#include "kmp.h"
+#include "omp.h"
 #include "perf_cnt.h"
 #include "printf.h"
 #include "riscv.h"
+// #include "snitch_cluster_global_interrupts.h" # what is Banshee equivalent?
```
colluca, line 34: Can we resolve this before merging?

target/snitch_cluster/sw/tests/banshee-apps.list
colluca, line 1: I like the alphabetic ordering, but perhaps we should enforce it, otherwise I fear it's not gonna last long :)

.github/workflows/ci.yml
colluca: You mean the Python script was not returning an non-zero exit code upon VCS failing? I think we should fix that instead of adding more logic

docs/ug/getting_started.md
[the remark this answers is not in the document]
colluca: My personal idea would be that if they choose not to use the container they should be able to figure it out themselves by following the `Dockerfile`. It is the most up-to-date, tested and detailed documentation you can get. Okay supporting partners, but that's what the Docker container is for.

target/snitch_cluster/README.md
colluca: Okay, so just the typo to fix then we can resolve this.

target/snitch_cluster/sw/runtime/banshee/src/snrt.h
GiannaP: This will need a discussion about the Banshee features with @SamuelRiedel I am not sure if Banshee currently has an interrupt model in Banshee.

target/snitch_cluster/sw/runtime/banshee/src/snitch_cluster_start.c
```diff
@@ -2,19 +2,20 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+// needs to be before #define SNRT_CRT0_EXIT
```
colluca, line 5: Are you sure? Doesn't it just have to be before `#include "start.c"`?

target/snitch_cluster/sw/runtime/banshee/src/snrt.h
colluca: Okay, but if there is no Banshee equivalent then we can just remove this line

PR #38, opened 2023-08-22
CSR dump

viv-eth:
This utility was inspired by MemPool to enable single cycle printf in RTL simulation.
This will **not** work on the actual hardware, but aids debugging. 

You have the option to interpret the register values as either 32-bit float values or 32-bit unsigned integers. Depending on what kind of data you want to display, use the `dump_float` or `dump_uint` utility, respectively. 

As an example, define `dump_float(<name>, <csr_reg_ID>` at the beginning of the file you want to debug. Afterward, you can dump the values to the output of the RTL simulation by calling `dump_<name>(<value>)`. 

You have to be careful when selecting the `csr_reg_ID` as it must **not** be an invalid register ID.

### DIFF
--- a/hw/snitch/src/snitch.sv
+++ b/hw/snitch/src/snitch.sv
@@ -2562,7 +2562,8 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
     // Display CSR write if the CSR does not exist
     if (!rst_i && csr_dump && inst_valid_o && inst_ready_i && !stall) begin
       // $timeformat(-9, 0, " ns", 0);
-      $display("[DUMP] %t Core %3d: 0x%3h = 0x%08h, %d, %f", $time, hart_id_i, inst_data_i[31:20], alu_result, alu_result, $bitstoshortreal(alu_result));
+      $display("[DUMP] %t Core %3d: 0x%3h = 0x%08h, %d, %f",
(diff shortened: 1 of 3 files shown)


hw/snitch/src/snitch.sv
```diff
@@ -2239,6 +2240,7 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
   // CSR logic
   always_comb begin
     csr_rvalue = '0;
+    csr_dump = '0;
```
fischeti, line 2243: use single bit assignment i.e. `1'b0`

hw/snitch/src/snitch.sv
```diff
@@ -2468,7 +2470,10 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
               if (!exception) fcsr_d = fcsr_t'(alu_result[9:0]);
             end else illegal_csr = 1'b1;
           end
-          default: csr_rvalue = '0;
+          default: begin
+            csr_rvalue = '0;
+            csr_dump = '1;
```
fischeti, line 2475: same as above

fischeti:
Otherwise LGTM. Thanks!

fischeti (approved):

colluca:
I think before we merge this we should discuss the purpose of it. I don't see a reason why we need this or why we should prefer this over the existing printf implementation.

PR #44, opened 2023-09-04
Revert "CSR dump"

viv-eth:
Reverts pulp-platform/snitch_cluster#38 because it is still open for discussion


colluca (changes requested):
There were changes to the container which are not being reverted here

PR #46, opened 2023-09-07
sw: Make compatible with non-12 LLVM versions

paulsc96:
This targets the work-in-progress occamy branch, *not* `main`.

Make software build system work with LLVM versions other than 12.0.1 by:

* Making LLVM lib paths adaptive to the version used
* Fixing GNU C language violations in `gemm.h`: `register` may not be used with arrays (this was never legal to begin with, but the frontend lets it slide and the backend only fatals if it actually attempts to map registers)


fischeti (approved):

JosseVanDelm:
Hi, this is quite exciting. 
Which versions of LLVM will be tested?
Will it be the vanilla LLVM version or a custom/pulp one?

Thanks. Keep up the good work!

PR #47, opened 2023-09-11
Various modifications from Occamy development

colluca:


paulsc96 (approved):
I can't speak much on SW stuff, but I don't mind merging this.

Maybe someone more familiar with the runtime and bundled apps can have another look.

viv-eth (approved):

PR #51, opened 2023-09-26
sw: Add sample Monte Carlo application

colluca:
When merging this we should add the following lines to the end of the commit message:

```
Co-authored-by: Fabio Cappellini <fcappellini@student.ethz.ch>
Co-authored-by: Hakim Filali <hfilali@student.ethz.ch>
```


fischeti (commented):
LGTM. Is the auto-merging still enabled in the repository? Because in that case it will be immediately merged when I approve this PR, and we cannot change the commit message to include the co-authors

PR #52, opened 2023-09-28
hw: Add optional cluster alias region

paulsc96:


hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -1246,6 +1289,8 @@ module snitch_cluster
   `ASSERT_INIT(CheckSuperBankFactor, (NrBanks % BanksPerSuperBank) == 0);
   // Check that the cluster base address aligns to the TCDMSize.
   `ASSERT(ClusterBaseAddrAlign, ((TCDMSize - 1) & cluster_base_addr_i) == 0)
+  // Check that the cluster alias address, if enabled, aligns to the TCDMSize.
+  `ASSERT(AliasRegionAddrAlign, ~AliasRegionEnable || ((TCDMSize - 1) & AliasRegionBase) == 0)
```
colluca, line 1293: Given that these are all constant parameters I believe `` `ASSERT_INIT `` would be more appropriate here

colluca (commented):
Otherwise looks good to me :)

Should I add a test to verify the functionality? This would require enabling the alias in the default cluster configuration, as the CI can currently test only one configuration. Otherwise we can extend the CI to test multiple configurations, which I think will be required in the long term, but we should discuss implementation details.

hw/snitch_cluster/src/snitch_cluster.sv
paulsc96: Sounds good, thanks for the commits :smiley:

colluca:
To keep track of the status, only Banshee tests are failing as this architectural feature is yet to be modeled in Banshee.

paulsc96:
As requested, this should be made configurable instead. Will look into that next.

colluca:
> As requested, this should be made configurable instead. Will look into that next.

After clarifying this, it shouldn't be necessary anymore.

fischeti (approved):
LGTM

PR #58, opened 2023-10-24
target/common: Scale verilator testbench time precision as picoseconds

### LINKED ISSUE #57: Verilator testbench: time is always 0 in DASM and perf results
I am currently following the tutorial for snitch cluster at https://pulp-platform.github.io/snitch_cluster/ug/tutorial.html and reached the debugging/benchmarking step. However, when trying to analyze the performance, it seems that the `tstart` and `tend` metrics are always 0. This value persists from the DASM file to the text, csv, and json results from the other steps for benchmarking. 

image

The command I ran: `bin/snitch_cluster.vlt sw/apps/blas/axpy/build/axpy.elf`
Resulting files is attached.
[trace_hart_00000000.dasm.txt](https://github.com/pulp-platform/snitch_cluster/files/13043772/trace_hart_00000000.dasm.txt)
[hart_00000000_perf.json](https://github.com/pulp-platform/snitch_cluster/files/13043779/hart_00000000_perf.json)
[perf.csv](https://github.com/pulp-platform/snitch_cluster/files/13043802/perf.csv)
[event.csv](https://github.com/pulp-platform/snitch_cluster/files/13043803/event.csv)

Running on docker on Linux, amd64: 
```bash
$ uname -a
Linux b0f761a4bb94 6.5.0-1004-oem #4-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 15 19:52:30 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

Tested running on
- axpy in tutorial
- blas/axpy from repo

Things I tried changing:
- Adding back the patch from https://github.com/pulp-platform/snitch/blob/d026f47843f0ea6c269244c4e6851e0e09141ec3/util/patches/riscv-isa-sim/fesrv.patch

Raragyay:
Fixes #57 

Currently, Verilator relies on the global function `sc_time_stamp` to implement the `$time` function in verilog. `$time` is used by the simulator's trace and the performance tooling in the repo to track performance. However, although the current implementation of the testbench increments the time counter on each cycle, it is still dividing this time counter by `1e9`. With the default assumption of 1 cycle taking 1ns and time precision of 1ps, this would imply that we are representing the `sim::TIME` variable as being in attoseconds (`1e18 attoseconds = 1 second`). This PR instead scales the time variable to return the expected picosecond precision. 

Reference example from Verilator 4.100 where no multiplication by `1e-9` takes place: https://github.com/verilator/verilator/blob/v4.100/examples/make_tracing_c/sim_main.cpp

Expected result: `time = cycle *1000 + 3000`. When generating traces in [gen_trace.py](https://github.com/pulp-platform/snitch_cluster/blob/main/util/trace/gen_trace.py#L501-L506) we assume that the time is in milliseconds, so it gives a good enough approximation. A next step could be to have wall time instead, but ultimately I think this leaves the verilator simulator in a better state than before. 

Example logs generated after the patch attached. 

[trace_hart_00000000.txt](https://github.com/pulp-platform/snitch_cluster/files/13166511/trace_hart_00000000.txt)
[perf.csv](https://github.com/pulp-platform/snitch_cluster/files/13166514/perf.csv)
[hart_00000000_perf.json](https://github.com/pulp-platform/snitch_cluster/files/13166515/hart_00000000_perf.json)
[event.csv](https://github.com/pulp-platform/snitch_cluster/files/13166516/event.csv)

### DIFF
--- a/target/common/test/verilator_lib.cc
+++ b/target/common/test/verilator_lib.cc
@@ -83,7 +83,7 @@ void Sim::main() {
 }  // namespace sim
 
 // Verilator callback to get the current time.
-double sc_time_stamp() { return sim::TIME * TIME_CYCLES_TO_TIMESTAMP; }
+double sc_time_stamp() { return sim::TIME * sim::TIME_CYCLES_TO_TIMESTAMP; }
 
 // DPI calls.
 void tb_memory_read(long long addr, int len, const svOpenArrayHandle data) {
--- a/target/common/test/verilator_lib.cc
+++ b/target/common/test/verilator_lib.cc
@@ -14,6 +14,11 @@ namespace sim {
 
 // Number of cycles between HTIF checks.
 const int HTIFTimeInterval = 200;
+
+// We want to return timestamp in picosecond accuracy, assuming that one cycle
+// takes 1ns Since 1 cycle takes 2 sim::TIME increments, scale by 500 to get
+// time = cycle * 1000 + <some constant>
+const int TIME_CYCLES_TO_TIMESTAMP = 500;
 void sim_thread_main(void *arg) { ((Sim *)arg)->main(); }
 
 // Sim time.
@@ -78,11 +83,7 @@ void Sim::main() {
 }  // namespace sim
 
 // Verilator callback to get the current time.
-double sc_time_stamp() { 
-    // We want to return timestamp in picosecond accuracy, assuming that one cycle takes 1ns
-    // Since 1 cycle takes 2 time increments, scale by 500 to get time = cycle * 1000 + <some constant>
-    return sim::TIME * 500; 
-}
+double sc_time_stamp() { return sim::TIME * TIME_CYCLES_TO_TIMESTAMP; }
 
 // DPI calls.
 void tb_memory_read(long long addr, int len, const svOpenArrayHandle data) {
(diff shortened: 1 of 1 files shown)


colluca:
Thanks for the contribution @Raragyay.

The RTL testbench uses a time precision of 1ps, hence the simulation times in the trace in picoseconds. We also generally assume a 1ns clock period. Ideally, we want to align the Verilator time to these default settings, i.e.:
```
time = 1000*cycle + <anything>
```

I would merge your contribution with these changes altogether.

colluca:
Hi @Raragyay,

Could you fix the formatting as suggested in the C linting action? https://github.com/pulp-platform/snitch_cluster/actions/runs/6641088274/job/18043051127?pr=58#step:4:27

Raragyay:
> Hi @Raragyay,
> 
> Could you fix the formatting as suggested in the C linting action? https://github.com/pulp-platform/snitch_cluster/actions/runs/6641088274/job/18043051127?pr=58#step:4:27

Will get onto it once I get back on dry land! Thank you for the reminder 🙂

colluca (approved):

colluca:
Thanks for the contribution @Raragyay!

PR #62, opened 2023-11-03
target/cluster: Expose cluster base ports

fischeti:
Make it possible to generate cluster wrappers that expose `cluster_base_addr_i` and `hart_base_id_i` for better multi-cluster support


colluca:
Looks good to me :)

Could you provide a PR to align/test this also in Occamy?

I see that the (now) legacy `tie_ports` parameter is used here:
https://github.com/pulp-platform/occamy/blob/0743f4eb5e71b77edef9199c42d874b907af7de4/util/occamygen/occamy.py#L59

fischeti:
I created a PR for the occamy repo https://github.com/pulp-platform/occamy/pull/11

Occamy currently uses the `occamy` branch of the snitch cluster. Should this branch be fast-forwarded once this PR is merged?

colluca (approved):

paulsc96:
Is there a reason we want to abandon the `tie_ports` parameter name and its polarity? I don't object, but it is a more disruptive change.

fischeti:
> Is there a reason we want to abandon the `tie_ports` parameter name and its polarity? I don't object, but it is a more disruptive change.

I mainly changed it since I thought is more descriptive this way. Plus, there is already the `sram_cfg_expose` parameter which essentially does the same thing. But I am also okay with keeping `tie_ports`

viv-eth:
Should we merge this or is the `tie_ports` still open?

paulsc96 (approved):

PR #64, opened 2023-11-11
target/common: Fix dasm traces truncation when using SnitchSim

Raragyay:
When using the `verification.py` library and `SnitchSim` to run a simulation then read results from the kernel execution, the IPC thread will crash the whole program once SnitchSim has finished issuing the read commands. This is because the current check for EOF in `ipc.cc` erroneously checks for EOF before the `fread` attempt, instead of after the `fread` attempt. See the following StackOverflow for how this is usually corrected:

https://stackoverflow.com/questions/5431941/why-is-while-feoffile-always-wrong

Because of this erroneous check, the last operation (which is usually a read memory operation) will be doubly executed, causing the attempt to write to and flush `rx` to fail, since `SnitchSim.py` will have already closed the pipe and directory. As such, the program crashes and the `.dasm` traces are never fully written out. 

This PR fixes the EOF check in `ipc.cc` to correctly check for EOF or errors after a read attempt instead of before.

### DIFF
--- a/target/common/test/ipc.cc
+++ b/target/common/test/ipc.cc
@@ -19,12 +19,16 @@ void* IpcIface::ipc_thread_handle(void* in) {
     // Handle commands
     ipc_op_t op;
 
-    while (!feof(tx)) {
-        uint8_t ret_value = fread(&op, sizeof(ipc_op_t), 1, tx);
-        if (ret_value != 1) {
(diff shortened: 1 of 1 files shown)


colluca (approved):

colluca:
Thanks a lot for your contribution @Raragyay!

PR #65, opened 2023-11-14
add chisel flow and sw test for block gemm

xiaoling-yi:
In this PR, we add the following:
* add chisel flow using bender
* add sw test for block gemm


xiaoling-yi:
Sorry, wrong RP!

PR #69, opened 2023-12-02
(Not a PR to main): Improve DMA tracer handling of 1d requests after 2d requests and scale units correctly

Raragyay:
This PR achieves two things:
- When a 1D DMA request is issued after a 2d request, the repetition count of the 2d request is still retained in the placeholder transaction. This causes the # bytes transferred tracking in `eval_dma_metrics` to omit generating a new DMA transfer placeholder after the 1D dma request is completed, leading to an IndexError  on `outst_transfers` when `exp_bursts` is accessed afterwards. 
- Bandwidth is stated to be calculated as bytes/cycle. However, the `start`/`end` metric currently reported in the DMA verilog `hw/future/src/dma/axi_dma_backend.sv` is measured in picoseconds (1000 * cycles + <constant>) as mentioned in https://github.com/pulp-platform/snitch_cluster/pull/58:
https://github.com/pulp-platform/snitch_cluster/blob/82233b608c239e49fdc7efaf9c0681aec8dd31d1/hw/future/src/dma/axi_dma_backend.sv#L338
as such, scale time down by 1000 to convert it into cycles unit.

### DIFF
--- a/util/trace/gen_trace.py
+++ b/util/trace/gen_trace.py
@@ -410,6 +410,7 @@ def update_dma(insn, extras, dma_trans):
             # Create new placeholder transaction
             dma_trans.append(dma_trans[-1].copy())
             dma_trans[-1].pop('size')
+            dma_trans[-1]['rep'] = 1
 
 
 def eval_dma_metrics(dma_trans, dma_trace):
--- a/util/trace/gen_trace.py
+++ b/util/trace/gen_trace.py
@@ -457,7 +457,7 @@ def eval_dma_metrics(dma_trans, dma_trace):
                         compl_transfers.append(compl_transfer)
         # Calculate bandwidth of individual transfers
         for transfer in compl_transfers:
-            transfer['cycles'] = transfer['end'] - transfer['start']
+            transfer['cycles'] = (transfer['end'] - transfer['start']) // 1000
             transfer['bw'] = transfer['bytes'] / transfer['cycles']
         # Calculate aggregate bandwidth: total number of bytes transferred while any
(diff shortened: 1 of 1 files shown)


colluca:
Thank you for the suggestions @Raragyay.

Regarding the issues you mentioned:
- 1D DMA transfers are initiated either by resetting the repetition count to 1 explicitly via DMREP, or by setting the enable_2d bit in the config field of DMCPY* to zero (see [documentation](https://pulp-platform.github.io/snitch_cluster/rm/custom_instructions.html)). In the latter case, your observation is correct. However, we wish to retain the value configured in the repetition counter for successive 2D transfers, so we cannot blindly reset the placeholder transaction, but we must instead parse the config field. 
- From the SV LRM: "The `$time` system function returns an integer that is a 64-bit time, scaled to the time unit of the module that invoked it." Verilator uses the default timescale of "1ps/1ps" when "timescale" does not occur before a given module.  With Questa, we explicitly override the timescale to use a time unit of 1ns.
To avoid parsing the traces differently based on the simulator used to produce them, I would align the Verilator time unit to 1ns, using the `--timescale 1ns/1ps` flag.

Both fixes are now implemented in https://github.com/pulp-platform/snitch_cluster/pull/71, so I would proceed to close this PR.

One final note: I tested the Verilator implementation with version 4.228, since the version we use by default (4.110) does not support `foreach` statements, and hence produces empty DMA traces.

Raragyay:
Thank you for considering my suggestions and clarifying the inner workings of DMA!

PR #70, opened 2023-12-05
treewide: Refactor Python simulation framework and fix process leakage

colluca:
- [x] Align CI and build scripts with SEPP package usage recommendations
- [x] Switch from Questa's two-step to the three-step flow: avoids re-optimizing on every simulation and the contention on the lock file by parallel simulations
- [x] Fix "Broken pipe" errors introduced by https://github.com/pulp-platform/snitch_cluster/pull/64 (went uncaught by CI)
- [x] Fix leakage of unterminated simulation processes and refactor Python simulation framework
- [x] [Enable colors](https://github.com/termcolor/termcolor/blob/main/src/termcolor/termcolor.py#L122C3-L123) in the Gitlab CI (`ansi` setting is not needed, but suggested [here](https://docs.gitlab.com/ee/ci/yaml/script.html#job-log-output-is-not-formatted-as-expected-or-contains-unexpected-characters))
- [x] Remove some Make targets which have already been promoted to `common.mk`


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: Build documentation
     runs-on: ubuntu-22.04
     container:
-      image: ghcr.io/pulp-platform/snitch_cluster:main
+      image: ghcr.io/pulp-platform/snitch_cluster:test-ci
```
paulsc96, line 18: TODO: Revert before merge

.github/workflows/ci.yml
```diff
@@ -29,45 +29,43 @@ jobs:
     name: Simulate SW on Snitch Cluster w/ Verilator
     runs-on: ubuntu-22.04
     container:
-      image: ghcr.io/pulp-platform/snitch_cluster:main
+      image: ghcr.io/pulp-platform/snitch_cluster:test-ci
```
paulsc96, line 32: TODO: revert before merge

.github/workflows/ci.yml
```diff
@@ -29,45 +29,43 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: 'recursive'
       - name: Build Software
         run: |
           bender vendor init
-          make -C target/snitch_cluster sw
+          make -C target/snitch_cluster sw -j
```
paulsc96, line 40: Parallel make is very different from serial invocation and can cause _spectacular_ issues when the targets are not properly written (I speak from experience). I trust that you know this and tested this well.

.github/workflows/ci.yml
```diff
@@ -29,45 +29,43 @@ jobs:
...
   # Build SW on Snitch Cluster w/ Banshee #
-  ############################################
+  #########################################
 
   sw-snitch-cluster-banshee:
     name: Simulate SW on Snitch Cluster w/ Banshee
     runs-on: ubuntu-22.04
     container:
-      image: ghcr.io/pulp-platform/snitch_cluster:main
+      image: ghcr.io/pulp-platform/snitch_cluster:test-ci
```
paulsc96, line 57: TODO: revert before merge

.github/workflows/ci.yml
paulsc96: This goes for all following cases, too.

sw/tests/while.c
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 ETH Zurich and University of Bologna.
```
paulsc96, line 1: This is a test for the new CI setup, correct? Do you intend to keep it?

target/common/common.mk
```diff
@@ -150,25 +161,30 @@ endef
...
+	@! grep -P "Errors: [1-9]*," $(dir $<)vopt.log
 	@mkdir -p $(dir $@)
 	@echo "#!/bin/bash" > $@
 	@echo 'binary=$$(realpath $$1)' >> $@
 	@echo 'mkdir -p $(LOGS_DIR)' >> $@
 	@echo 'echo $$binary > $(LOGS_DIR)/.rtlbinary' >> $@
 	@echo '${VSIM} +permissive ${VSIM_FLAGS} $$3 -work ${MKFILE_DIR}/${VSIM_BUILDDIR} -c \
 				-ldflags "-Wl,-rpath,${FESVR}/lib -L${FESVR}/lib -lfesvr -lutil" \
-				$1 +permissive-off ++$$binary ++$$2' >> $@
+				$1_opt +permissive-off ++$$binary ++$$2' >> $@
```
paulsc96, line 179: This is legal, but in such cases I prefer explicit variable name delimitation (`$(1)_opt`).

Or at least I assume this is a make call argument (for bash it'd be `${1}_opt`)

util/sim/sim_utils.py
```diff
@@ -0,0 +1,166 @@
...
+                    sim.print_status()
+                else:
+                    failed_sims.append(sim)
+                    sim.print_log()
+                    sim.print_status()
+                    # If in early-exit mode, terminate as soon as any simulation fails
+                    if early_exit:
+                        early_exit_requested = True
+                        break
+            time.sleep(0.2)
```
paulsc96, line 156: Maybe it's worth making this a command line parameter

paulsc96 (changes requested):
The process management is still a bit hard to follow and the newly introduced use of make job servers on existing targets makes me nervous. That said, if you are confident it works, let's merge it after some cleanup (see comments).

.github/workflows/ci.yml
colluca: I tested it just by invoking the targets, which the CI also does. But I will drop the relevant commit then; the speedup on the overall CI time should not be so significant anyways.

target/common/common.mk
colluca: Good catch yes, that makes sense. (it is a make call argument)

colluca:
Ready to be merged after your approval @paulsc96. The CI should pass on main (as the Docker container will be updated).

paulsc96 (approved):

PR #74, opened 2023-12-19
Modify data index generation

xiaoling-yi:
This PR solves this question: https://github.com/KULeuven-MICAS/snitch_cluster/issues/54


xiaoling-yi:
Soorrry!!

Issue #75, opened 2023-12-21
Support Verilator 5


RickLuiken:
Following from the [discussion in the fork from KU Leuven](https://github.com/KULeuven-MICAS/snitch_cluster/issues/65). I would like to have support for Verilator 5. Verilator 5 has better performance, support for timing constructs and support for assertions. It would also allow projects using the snitch cluster to use verilator 5. I think in general it would be useful to support more recent versions of tools, as this makes the upgrading process later down the line a lot easier (make small steps instead of big ones).

Upgrading to verilator 5 would involve the following steps:
- Add the `-no-timing` flag to the current tests, which disables the new timing features in verilator 5
- Add `verilated_threads.o` to the verilator build targets **OR** compile both the systemverilog and cpp sources using verilator, so verilator can manage which verilator files should be included

If you have any questions or comments, please let me know!

colluca:
Dear @RickLuiken,

If you could provide a PR for this feature targeting Verilator 5.006, it would be well accepted and we would be willing to review and merge it.

Unfortunately we do not have enough resources to implement this feature ourselves at the moment.

RickLuiken:
Great, I will provide the PR after the christmas break!

colluca:
This is currently undergoing testing in https://github.com/pulp-platform/snitch_cluster/pull/158.

colluca:
Merged in https://github.com/pulp-platform/snitch_cluster/pull/158.

PR #76, opened 2024-01-04
Switch to Verilator 5 and update other dependencies

### LINKED ISSUE #6: Different Verible version in Docker container and Github CI linting stage

RickLuiken:
Fixes #6 and #75.

I changed the way the Makefile works to compile both the sv and c++ files with verilator. Moreover, I did the following:
- Added a `no-timing` flag to the verilator command
- Changed the c++ std from c++14 to c++20 (recommended by verilator)
- Changed the spike fesvr version to a more recent one to support GCC13 (see [this PR](https://github.com/riscv-software-src/riscv-isa-sim/pull/1284)). Unfortunately, this newer version requires at least c++17, which is not compatible with verilator 4. Therefore, verilator 4 does not work with this PR!
- Updated the dockerfile. It now uses ubuntu 22.04. It builds verilator 5.006 from the source and now uses the same verible version as the CI.

To test it, I did the following:
- [x] Run verilator tests using docker
- [x] Run banshee tests using docker

I do not have access to modelsim, so I couldn't test that.

If you have any comments/questions, please let me know!

### DIFF
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -31,9 +31,6 @@ PERF_CSV_PY      ?= $(UTIL_DIR)/trace/perf_csv.py
 LAYOUT_EVENTS_PY ?= $(UTIL_DIR)/trace/layout_events.py
 EVENTVIS_PY      ?= $(UTIL_DIR)/trace/eventvis.py
 
-VERILATOR_ROOT ?= $(dir $(shell $(VERILATOR_SEPP) which verilator))..
-VLT_ROOT       ?= ${VERILATOR_ROOT}
-
 MATCH_END := '/+incdir+/ s/$$/\/*\/*/'
 MATCH_BGN := 's/+incdir+//g'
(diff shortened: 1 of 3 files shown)


RickLuiken:
Hi, just to let you know, the CI failed because it uses the old container with verilator 4. It should work if the new container is used, but I don't know how to test this in the CI. Could you help me with this?

colluca:
Thanks for the contribution @RickLuiken.

We currently have an issue with the CI which we are solving. After that I will try it out with the Docker container update and let you know.

Best,
Luca

colluca:
Hi @RickLuiken, once again thanks for the contribution. I reduced the changes and split them into two separate PRs, one of which has already been merged.

I will close this PR in favour of https://github.com/pulp-platform/snitch_cluster/pull/158 where we will follow on the  upgrade to Verilator 5.

Best,
Luca

PR #82, opened 2024-01-15
`snrt_global_barrier` requires additional `snrt_cluster_hw_barrier`

rogerbarton:
An additional `snrt_cluster_hw_barrier` is required before the DM cores synchronize. 

Currently, the DM cores do not know if the other compute cores have arrived at the barrier. For example if a compute core in cluster 0 is late arriving at the barrier, other clusters will not wait for this core.


fischeti (approved):
That makes sense. Thanks!

colluca (approved):

PR #83, opened 2024-01-19
sw: Add GEMM baselines without SSR & frep extensions

fischeti:
To compare the performance of the `Xssr` and `Xfrep` extensions, this PR adds baseline (optimized) GEMM kernels.

### DIFF
--- a/sw/blas/gemm/src/gemm.h
+++ b/sw/blas/gemm/src/gemm.h
@@ -1217,9 +1217,9 @@ void sc_st_gemm(precision_t prec, uint32_t expand, uint32_t setup_ssr,
             case FP64:
                 if (baseline) {
                     gemm_fp64_naive(frac_m, n, k, (double*)a + offsetA,
-                                       lda_strided, transa, (double*)b, ldb,
-                                       transb, (double*)c + offsetC,
-                                       ldc_strided, (double)beta);
+                                    lda_strided, transa, (double*)b, ldb,
+                                    transb, (double*)c + offsetC, ldc_strided,
+                                    (double)beta);
                 } else {
                     gemm_fp64_opt(frac_m, n, k, (double*)a + offsetA,
                                   lda_strided, transa, (double*)b, ldb, transb,
@@ -1230,8 +1230,8 @@ void sc_st_gemm(precision_t prec, uint32_t expand, uint32_t setup_ssr,
             case FP32:
                 if (baseline) {
                     gemm_fp32_baseline(frac_m, n, k, (float*)a + offsetA,
-                                       lda_strided, (float*)b, ldb, (float*)c + offsetC, ldc_strided,
-                                       beta);
+                                       lda_strided, (float*)b, ldb,
+                                       (float*)c + offsetC, ldc_strided, beta);
                 } else {
                     gemm_fp32_opt(frac_m, n, k, (float*)a + offsetA,
                                   lda_strided, (float*)b, ldb,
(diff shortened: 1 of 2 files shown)


sw/blas/gemm/verify.py
```diff
@@ -55,8 +59,8 @@ def main():
     c_golden = golden_model(1, a, b, beta, c).flatten()
 
     absolute_err = np.absolute(c_golden - c_actual)
-    fail = np.any(absolute_err > ERR_THRESHOLD)
-    if (fail):
+    fail = np.any(absolute_err > ERR_THRESHOLD[prec]) or np.isnan(c_actual).any()
```
colluca, line 62: Good catch. I guess this might be needed also in the other kernels' verification scripts.
Should we define a reusable function in `data_utils.py`?

sw/blas/gemm/verify.py
```diff
@@ -10,23 +10,27 @@
...
 
-sys.path.append(str(Path(__file__).parent / '../../../util/sim/'))
+sys.path.append(str(Path(__file__).parent / "../../../util/sim/"))
 import verification  # noqa: E402
 from elf import Elf  # noqa: E402
 from data_utils import from_buffer, ctype_from_precision_t  # noqa: E402
 
 
-ERR_THRESHOLD = 0.001
+ERR_THRESHOLD = {8: 1e-6, 4: 1e-6, 2: 1e-2, 1: 1e-1}
```
colluca, line 19: Can the absolute error actually be this large when using fp16 and fp8?

sw/blas/gemm/src/gemm.h
```diff
@@ -35,9 +35,9 @@ static inline double multiply_opt(double multiplicand, double multiplier) {
         return 0;
 }
 
-void gemm_fp32_baseline(uint32_t M, uint32_t N, uint32_t K, float* A,
-                        uint32_t ldA, uint32_t ta, float* B, uint32_t ldB,
-                        uint32_t tb, float* C, uint32_t ldC, float BETA) {
+void gemm_fp32_naive(uint32_t M, uint32_t N, uint32_t K, float* A, uint32_t ldA,
```
colluca, line 38: What is the difference between "naive" and "baseline" variants?

sw/blas/gemm/verify.py
fischeti: That would be a good idea yes. I will add that

sw/blas/gemm/verify.py
fischeti: It really depends on the problem size. The larger it is, the higher is the chance that you have some kind of catastrophic cancellation or saturation.

The error threshold for fp8 might even be higher. The problem is that it is almost impossible to verify fp8 kernels. fp8 is not a standard and lacks support in numpy, and we are also missing tracing of fp8 values currently, which makes it quite hard to debug and verify.

sw/blas/gemm/src/gemm.h
fischeti: The idea is that the baseline kernels should be a "fair" baseline against the SSR/frep extensions i.e. the best you can get out of Snitch without using the ISA extensions.

fischeti:
@viv-eth @colluca `flash_attention2` and `softmax` return NaNs, which was not captured before. Do you know what is going on?

fischeti:
> @viv-eth @colluca `flash_attention2` and `softmax` return NaNs, which was not captured before. Do you know what is going on?

I opened a new draft PR https://github.com/pulp-platform/snitch_cluster/pull/96 to address this. In the meantime we can merge this one once the CI passes

sw/blas/gemm/verify.py
```diff
@@ -10,23 +10,27 @@
...
+
+NP_DTYPE = {8: np.float64, 4: np.float32, 2: np.float16, 1: np.uint8}
```
colluca, line 21: This is not used, correct?

colluca (approved):

PR #87, opened 2024-01-30
DNN kernels to support GPT decoder models and additional utilities

viv-eth:
The SW libraries have been restructured into separate folders to ease data generation and verification (Occamy only). 

This PR implements the following kernels:

1. Multi-cluster GEMM with tiling + verification
2. Matrix concatenation + verification
3. FlashAttention-2 
4. Fused linear and concatenation layer with logarithmic reduction + verification
5. i-GeLu activation function + verification
6. LayerNorm 

The following utilities have been added:

1. Safe float/integer casts to ensure consistency 
2. Convenient 2D DMA transfer functions
3. Global Reduction function for binary tree reduction across multiple clusters

⚠️Currently dummy functions are used for the exponential due to a HW bug in the FPU and integer core synchronization. This will be **reverted** as soon as the feature is implemented.


colluca:
If we cherry-pick 18554fc57d8f4ecd424b222c4fcb94a784a6cbca and 420dfb3cc2f2478599d89f8c590a686e9564fc0f we can also add LayerNorm to the CI, while we wait for the release of the `-mno-fdiv` supporting toolchain.

fischeti:
Is this now a draft or not?

colluca:
> ⚠️Currently dummy functions are used for the exponential due to a HW bug in the FPU and integer core synchronization. This will be **reverted** as soon as the feature is implemented.

What do you mean "dummy" functions? I did implement correct versions in the math library, did you cherry-pick those commits?

sw/dnn/layernorm/verify.py
```diff
@@ -0,0 +1,86 @@
...
+    4: '32',
+    2: '16',
+    1: '8'
+}
+
+NUMPY_T = {
+    '64': np.float64,
+    '32': np.float32,
+    '16': np.float16
+}
```
fischeti, line 33: We could also move those to `data_utils`. Since they are used in every verify script

sw/dnn/layernorm/verify.py
```diff
@@ -0,0 +1,86 @@
...
+        'dtype': 'I'
+    }
+    layer = bytes_to_struct(elf.get_symbol_contents('layer'), layer_struct)
+    batch_size = layer['batch_size']
+    seq_len = layer['seq_len']
+    embeddings = layer['embeddings']
+    eps = layer['eps']
+    prec = PRECISION_T[layer['dtype']]
+
+    ifmap = np.array(bytes_to_float(elf.get_symbol_contents('ifmap'), prec), dtype=NUMPY_T[prec])
```
fischeti, line 68: In general, we could replace this with [`np.frombuffer`](https://numpy.org/doc/stable/reference/generated/numpy.frombuffer.html) in all verify scripts. See [here](https://github.com/pulp-platform/snitch_cluster/blob/9b6cd8de64212661c5fc3bb39dc952ce159980d9/sw/blas/gemm/verify.py#L57) for an example

target/snitch_cluster/cfg/divsqrt.hjson
fischeti, line 1: What's the difference between the `divsqrt` and `fdiv` configuration?

target/snitch_cluster/Makefile
```diff
@@ -53,7 +53,7 @@ VLT_AR = ${VLT_BUILDDIR}/Vtestharness__ALL.a
 # (LRU) config, all targets depending on the configuration file have
 # to be rebuilt. This file is used to express this condition as a
 # prerequisite for other rules.
-DEFAULT_CFG = cfg/default.hjson
+DEFAULT_CFG = cfg/divsqrt.hjson
```
fischeti, line 56: I am not sure if we want to enable the Div/Sqrt unit by default

target/snitch_cluster/cfg/divsqrt.hjson
```diff
@@ -0,0 +1,127 @@
...
+        num_fp_outstanding_mem: 4,
+        num_sequencer_instructions: 16,
+        num_dtlb_entries: 1,
+        num_itlb_entries: 1,
+        // Enable division/square root unit
+        Xdiv_sqrt: true,
+    },
+    dma_core_template: {
+        isa: "rv32imafd",
+        Xdiv_sqrt: true,
```
fischeti, line 108: Does the DMA core really need a Div/Sqrt Unit? The only reason it even has an FPU is because we ran into issues where the compiler might insert stack operations to store floating-point registers to the stack, which caused an exception in the DMA core, when it was configured without an FPU

target/snitch_cluster/sw/apps/Makefile
```diff
@@ -9,14 +9,17 @@ SUBDIRS += nop
 SUBDIRS += blas/axpy
 SUBDIRS += blas/gemm
 SUBDIRS += dnn/batchnorm
-SUBDIRS += dnn/conv2d
-SUBDIRS += dnn/fusedconv
+# SUBDIRS += dnn/conv2d
+# SUBDIRS += dnn/fusedconv
```
fischeti, line 13: Why are they not compiled?

util/sim/data_utils.py
```diff
@@ -6,15 +6,47 @@
...
 
 
+def floating_point_torch_type(precision):
+    prec_to_torch_type_map = {
+        '64': torch.float64,
+        '32': torch.float32,
+        '16': torch.float16,
+        '8': None
+    }
+    return prec_to_torch_type_map[precision]
```
fischeti, line 27: There is no need to wrap the dict as a function. We could declare `TORCH_DTYPE`, `NP_DTYPE` etc. as dicts here and import it in all the verify scripts

sw/blas/gemm/src/main.c
```diff
@@ -12,74 +13,26 @@
...
 int main() {
+    int retcode =
+        gemm(dtype_size, expand, 1, parallelize_m, parallelize_k, m_tiles,
+             n_tiles, k_tiles, 1, 1, 1, TA, TB, M, N, K, 1, a, b, BETA, c);
+
+    snrt_cluster_hw_barrier();
+
+// TODO: currently only works for single cluster otherwise need to
+//       synchronize all cores here
+#ifdef BIST
```
fischeti, line 27: `BIST` is undefined above

sw/blas/gemm/src/gemm.h
```diff
@@ -73,8 +237,21 @@ void gemm_fp64_opt(uint32_t M, uint32_t N, uint32_t K, double* A, uint32_t ldA,
     // Unrolling factor of most inner loop.
     // Should be at least as high as the FMA delay
     // for maximum utilization
-    const uint32_t unroll = 8;
-
+    // const uint32_t unroll = 8;
+    const uint32_t unroll = 4;
```
fischeti, line 241: What's the reason to decrease the unrolling? I tried it myself once, since the FPU latency should now be low enough to prevent RAW conflicts, but I remember that it reduced the FPU utilization by a couple of percent, which is why I decided to keep it at 8.

.github/workflows/lint.yml
```diff
@@ -117,6 +117,7 @@ jobs:
         with:
           flake8-version: "6.0.0"
           max-line-length: "100"
+          exclude: "target/snitch_cluster/sw/apps/dnn/datagen.py"
```
fischeti, line 120: What's the problem with that one? it doesn't report any linting errors for me

fischeti (changes requested):

sw/dnn/layernorm/verify.py
viv-eth: I am moving the `PRECISION_T` and `NUMPY_T`, but not the `ERR_THRESHOLD`, since it is different in some files, e.g. for `GELU`

sw/dnn/layernorm/verify.py
viv-eth: Leaving this to @colluca

target/snitch_cluster/cfg/divsqrt.hjson
viv-eth: Different timing parameters. @colluca which one should we keep? Currently in the tests it is the `FDIV` one

target/snitch_cluster/Makefile
viv-eth: @colluca leaving this to you.

target/snitch_cluster/cfg/divsqrt.hjson
viv-eth: @colluca Should we leave it until we have the new toolchain? Or the HW fixed?

util/sim/data_utils.py
viv-eth: @colluca

target/snitch_cluster/Makefile
colluca: True, the default should remain `default.hjson`. I will fix this.

target/snitch_cluster/cfg/divsqrt.hjson
colluca: For the same reason you mentioned above, i.e. the compiler not being aware of the heterogeneity of the cores' ISAs, I would enable it on all cores. Once the `-mno-fdiv` supporting toolchain is released we can get rid of the `divsqrt.hjson` configuration altogether.

sw/dnn/layernorm/verify.py
colluca: Saw this, definitely a good idea to use the numpy functions. I would however wrap them in our own custom function, such that we can handle `fp8` precision with the same function as well.

util/sim/data_utils.py
colluca: True, I just prefer functions in general for extensibility, we could e.g. add type-checking to it.

colluca (changes requested):

sw/blas/gemm/src/gemm.h
```diff
@@ -5,11 +5,13 @@
 // Author: Tim Fischer <fischeti@iis.ee.ethz.ch>
 //         Luca Bertaccini <lbertaccini@iis.ee.ethz.ch>
 //         Luca Colagrande <colluca@iis.ee.ethz.ch>
+//         Viviane Potocnik <vivianep@iis.ee.ethz.ch>
 
 #include <stdint.h>
 
 #include "snrt.h"
 
+#define BASELINE 1
```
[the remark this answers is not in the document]
fischeti: My suggestion would be to define another parameter in the `param.hjson` file for this instead of using defines. This would make profiling of kernels easier. I started this in another PR https://github.com/pulp-platform/snitch_cluster/pull/83

target/snitch_cluster/sw/apps/Makefile
viv-eth: conv2d fixed

sw/blas/gemm/data/params.hjson
```diff
@@ -5,12 +5,17 @@
...
-    M: 192,
-    N: 16,
-    K: 16,
+    M: 32,
+    N: 4,
+    K: 4,
     beta: 0,
     ta: false,
-    tb: true, // must be true for SIMD
+    tb: false, // must be true for SIMD
```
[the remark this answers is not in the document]
colluca: Update: CI doesn't pass when setting this to `true`

sw/dnn/linear/src/linear.h
```diff
@@ -127,33 +126,4 @@ static inline void linear_layer(const linear_layer_t *l) {
     }
 
     snrt_cluster_hw_barrier();
-
-    if (snrt_is_dm_core()) {
-        snrt_dma_txid_t txid_result = snrt_dma_start_2d(
-            result, l->result, l->CH * sizeof(float), l->CH * sizeof(float),
-            l->CH * sizeof(float), l->CO * sizeof(float));
-        snrt_dma_wait_all();
-    }
```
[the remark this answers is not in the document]
colluca: The kernel appears to be severely broken. This condition always evaluates to 0, and when changing it to a reasonable condition, the simulation raises `RegWriteKnown` errors: https://github.com/pulp-platform/snitch_cluster/blob/ae02a03d7d401ed486dac80a3a1d77e1a89c395d/sw/dnn/src/linear.h#L107C9-L107C54
Removing the linear layer altogether

sw/dnn/softmax/src/softmax.h
```diff
@@ -6,41 +6,58 @@
...
+    float fx = 0.0f;
+    union {
+        float f;
+        int i;
+    } v = {x};
+    v.i = (1 << 23) * (x / logf(2)) + 0x3f800000;
+    fx = v.f;
+    return fx;
+}
+
```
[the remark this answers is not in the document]
colluca: The `expf` function in the math library works, so we are removing these.

colluca (approved):

PR #99, opened 2024-02-16
target: Improve reusability of Questa Make targets

colluca:
- [x] Improves reusability of Questa Make targets, by introducing the `$(TARGET)` variable abstraction and moving the targets to a separate Make fragment
- [x] Adds a dependency on the `Bender.lock` file to `compile.vsim.tcl`


fischeti (approved):
LGTM!

Issue #105, opened 2024-02-24
Bug in GEMM FP32 baseline kernel


viv-eth:
When used in the FlashAttention-2 layer the GEMM FP32 kernel yields wrong results by orders of magnitude

colluca:
With what parameters do you observe these errors?

fischeti:
It's because of this:
https://github.com/pulp-platform/snitch_cluster/blob/d325300c8c2f6dd6b81b6612428e787adf133464/sw/blas/gemm/src/gemm.h#L1381-L1387
Matrix B **needs** to be transposed for SIMD kernels. Now it just immediately returns for all FP32 gemm kernels. It would be nice to have proper C assertions there instead of `return -1`. I tried that once. I tried that once, but ran into some issues with newlib.

PR #107, opened 2024-02-26
gemm: Test multiple configurations

fischeti:
Adresses #105 

The GEMM kernel previously checked for unsupported transpositions of matrices and returned a non-zero exitcode. However, the assertions were wrong. 

This PR removes the assertions from the GEMM kernel and moves them into the `datagen` script, where they should be easier to detect.


fischeti:
@viv-eth Can you check if that solves your issue?

colluca (approved):

colluca:
I added additional commits on top of this PR, to test multiple GEMM configurations:
- [x] Add assertions to the GEMM datagen script, to raise errors on configurations which are known to be not supported.
- [x] Split the `gemm.h` file into multiple files (by precision) to make it more manageable.
- [x] Correct a bug in `dump_results_to_csv()` when called not on a Numpy array (e.g. a list of FP8 values).
- [x] Move reusable functionality out of `run.py` script, so that other testing scripts can be developed deriving from those functionalities.
- [x] Allow defining arbitrary test names in the testlist specification file, rather than inferring them from the software binary. Allows giving unique names to different tests, despite the name of the binary (not the full path) is the same.
- [x] Allow overriding data header and configuration paths (GEMM only) from the environment to build different configurations.
- [x] Allow overriding SW build directory from the environment for the same reason as above.
- [x] Add PHONY target based on app name as an alias for the ELF target. Simplifies typing make commands.
- [x] Create a script (based on `run.py` functionality) to build and test multiple GEMM configurations.

PR #112, opened 2024-03-12
Replace snitch_icache with cluster_icache dependency

micprog:
In order to keep a single common source for IPs, the snitch_cluster instruction cache was extracted into its own repository. This repository was updated with features from Mempool, allowing reuse of common features, improvements, and bugfixes. This PR updates the icache to point to this new common source through a bender dependency.


fischeti (approved):

colluca:
Thanks @micprog!

LGTM, only thing I would change concerns the CI fix. Since https://github.com/pulp-platform/common_cells/pull/217 was merged to master already, I would temporarily use that hash in the `Bender.yml`, so we can then keep the `common_cells` assertions on with vsim. Later we can bump to v1.33.1. Alternatively, we can wait for the next release altogether (should not be delayed for long anyways).

Let me know what you think, I can implement the remaining changes and merge.

micprog:
I think waiting shortly for the common_cells release would make sense to keep consistency and maximum testability. Separating the CI fixes from this PR would also be valuable as a separation of concerns.

colluca (approved):

SamuelRiedel (approved):

PR #113, opened 2024-03-12
Add optimized LayerNorm kernels

viv-eth:
This PR adds the MiniFloat versions (Fp16 and FP8) of the LayerNorm kernel.


sw/dnn/layernorm/src/layernorm.h
```diff
@@ -246,26 +250,493 @@ static inline void layernorm_fp32_opt(float *input, float *output,
...
+                    "vfmul.b ft1, ft2, %[mean_reg] \n"
+                    : [ mean_reg ] "+f"(mean_reg.f64)
+                    : [ n_frep ] "r"(n_frep - 1)
+                    : "ft0", "ft1", "ft2");
+                snrt_ssr_disable();
+            }
+        }
+    }
+}
 // Tiles the seq_len axis (assumes seq_len is an integer multiple of n_tiles)
```
colluca, line 691: Just a style note: missing a a new line here

sw/dnn/layernorm/src/layernorm.h
```diff
@@ -246,26 +250,493 @@ static inline void layernorm_fp32_opt(float *input, float *output,
...
+    uint32_t data_type_size;
+    if (l.dtype == FP32) {
+        data_type_size = sizeof(float);
+    } else if (l.dtype == FP16) {
+        data_type_size = sizeof(__fp16);
+    } else if (l.dtype == FP8) {
+        data_type_size = sizeof(char);
+    } else {
+        printf("Unsupported data type\n");
+    }
```
colluca, line 704: The value of the enumeration type is actually already the `data_type_size`, so you don't need this here. See https://github.com/pulp-platform/snitch_cluster/blob/04f80a17905d06774d378ad9c932968d5c790090/sw/dnn/src/dnn.h#L13

sw/dnn/layernorm/src/layernorm.h
```diff
@@ -246,26 +250,493 @@ static inline void layernorm_fp32_opt(float *input, float *output,
...
 static inline void layernorm_layer(layernorm_layer_t l) {
+    uint32_t data_type_size;
+    if (l.dtype == FP32) {
+        data_type_size = sizeof(float);
+    } else if (l.dtype == FP16) {
+        data_type_size = sizeof(__fp16);
+    } else if (l.dtype == FP8) {
+        data_type_size = sizeof(char);
+    } else {
+        printf("Unsupported data type\n");
```
colluca, line 703: Add FP64 not supported to `validate_config()`

sw/dnn/layernorm/src/layernorm.h
```diff
@@ -246,26 +250,493 @@ static inline void layernorm_fp32_opt(float *input, float *output,
...
+        remote_ifmap = (__fp16 *)l.ifmap;
+        remote_ofmap = (__fp16 *)l.ofmap;
+    } else if (l.dtype == FP8) {
+        local_itile = (char *)snrt_l1_next();
+        local_otile = local_itile + tile_size;
+        remote_ifmap = (char *)l.ifmap;
+        remote_ofmap = (char *)l.ofmap;
+    } else {
+        printf("Unsupported data type\n");
+    }
```
colluca, line 739: Multiple cases are not needed, since the LHS pointers are all of void type.

sw/dnn/layernorm/src/layernorm.h
```diff
@@ -294,11 +765,25 @@ static inline void layernorm_layer(layernorm_layer_t l) {
...
+            } else if (l.dtype == FP16) {
+                layernorm_fp16_opt(local_itile, local_otile, l.batch_size,
+                                   tile_seq_len, l.embeddings, l.eps);
+            } else if (l.dtype == FP8) {
+                layernorm_fp8_opt(local_itile, local_otile, l.batch_size,
+                                  tile_seq_len, l.embeddings, l.eps);
+            } else {
+                printf("Unsupported data type\n");
+            }
         }
```
colluca, line 787: For consistency with the GEMM implementation I would first test dtype with a concise switch case statement, and internally for each precision select between the different implementations

sw/dnn/src/dnn.h
```diff
@@ -10,7 +10,7 @@
 // TODO: move this definition to Snitch math library to solve problem
 #ifndef PRECISION_T
 #define PRECISION_T
-typedef enum { FP64 = 8, FP32 = 4, FP16 = 2, FP8 = 1 } precision_t;
+typedef enum { FP64 = 8, FP32 = 4, FP16 = 2, FP8 = 1, FP8ALT = 1 } precision_t;
```
colluca, line 13: Not used, why do we need to add this?

util/sim/data_utils.py
```diff
@@ -32,7 +32,7 @@ def emit_license():
 # Enum value can be a string or an integer, this function uniformizes the result to integers only
 def _integer_precision_t(prec):
     if isinstance(prec, str):
-        return {'FP64': 8, 'FP32': 4, 'FP16': 2, 'FP8': 1}[prec]
+        return {'FP64': 8, 'FP32': 4, 'FP16': 2, 'FP8': 1, 'FP8ALT': 1}[prec]
```
colluca, line 35: Not used, why do we need to add this?

colluca (changes requested):
Can we also add verification for the different precisions in the CI, e.g. as the GEMM kernel has?

colluca (approved):

PR #114, opened 2024-03-12
sim: Support non-default logs directory

fischeti:
Mostly useful for multiple measurement simulations in parallel without overwritting the traces needed for performance measurements. Can also be used in conjunction with custom build directory, to build multiple similar binaries (e.g. with different problem dimensions) 

To be used like this:

```bash
# build app in separate folder (optional)
make sw APP_BUILDDIR=my_build
# Run the simulation
./bin/snitch_cluster.vsim sw/path/to/my_build/app.elf "" +LOGS_DIR=my_logs
# make the traces
make traces LOGS_DIR=my_logs
```

### DIFF
--- a/hw/snitch_cluster/src/snitch_cc.sv
+++ b/hw/snitch_cluster/src/snitch_cc.sv
@@ -855,6 +855,7 @@ module snitch_cc #(
   // pragma translate_off
   int f;
   string fn;
+  string logs_dir, mkdir_cmd;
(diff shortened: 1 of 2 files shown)


fischeti:
Closing this since the better way to do it is to run the simulation from another subdirectory i.e.

```bash
mkdir my_run && cd my_run
../bin/snitch_cluster.vsim ../sw/path/to/my_elf.elf
```

PR #115, opened 2024-03-13
`gen_trace.py`: Represent SIMD operands as vectors

colluca:
### Contributions
- Extends the tracer to decode SIMD operands as vectors.
- Redefines the `flt_fmt()` function which was buggy: the `math.frexp()` return values were inverted and also misinterpreted! These are base 2 mantissa and exponents.
- Fixes a bug in `main()` where `fpr_wb_info` was used in place of `gpr_wb_info`.
- Adds docstrings and comments to several functions.

### TODO
- [x] Double-check that non-SIMD operations preserve the previous behaviour
- [x] Implementation is quite slow: continuously reads the CSV file and iterates through all rows. This can be improved simply by caching.


colluca:
It looks like the previous behaviour is not always maintained, e.g. on the scalar instruction `fdiv.s` we have SIMD register representations in the trace:
```
#; main (main.c:13:5)
#;   flashattention_2_layer (flashattention_2.h:51:13)
#;     flashattention_2_fp32 (flashattention_2_fp32.h:245:49)
#;       O_fa[row_idx * d + col_idx] /= l_i[row_idx];
#;                                   ^
                      0x80004a78 fdiv.s  ft0, ft1, ft0          #; ft1  = [nan, 101.32888793945312], ft0  = [nan, 8.307751655578613]
```

fischeti (approved):
This is very nice, thanks a lot! Is the CSV file auto-generated or was this assembled by hand?

colluca:
> This is very nice, thanks a lot! Is the CSV file auto-generated or was this assembled by hand?

This was assembled by hand by @viv-eth, with some corrections made by me, but it has not been exhaustively verified. Indeed, auto-generating this would be the best approach, but we didn't see an easy solution. Of course, if you see a way this could be done, it's worth considering it.

PR #121, opened 2024-03-19
GEMM: Add switch cases for different GEMM flavors.  

viv-eth:
The new `implementation_t` let's you now select which kernel variant you want to run. The supported flavors include: 

1. `BASELINE` : Assembly-based baseline kernel (FP32, FP16, FP8)
2. `NAIVE` : Simple for-loops for computing the GEMM (FP64, FP32, FP8)
3. `NAIVE_UNROLLED` : for-loops with loop unrolling (FP32)
4. `OPT` : Optimized kernels leveraging SSRs and FREP (FP64, FP32, FP16)
5. `OPT_EX` : Optimized low-precision kernels with expanding to the next higher precision (FP16, FP8)

### DIFF
--- a/sw/blas/gemm/scripts/datagen.py
+++ b/sw/blas/gemm/scripts/datagen.py
@@ -5,6 +5,7 @@
 
 # Authors: Tim Fischer     <fischeti@iis.ee.ethz.ch>
 #          Luca Bertaccini <lbertaccini@iis.ee.ethz.ch>
+#          Viviane Potocnik <vivianep@iis.ee.ethz.ch>
 
 import numpy as np
 import argparse
@@ -31,8 +32,8 @@ def golden_model(alpha, a, b, beta, c):
     return alpha * np.matmul(a, b) + beta * c
 
 
-def validate_config(prec, parallelize_m, parallelize_k, m_tiles, n_tiles, k_tiles, ta, tb, M, N, K,
-                    baseline, beta, **kwargs):
+def validate_config(prec, implementation, parallelize_m, parallelize_k, m_tiles, n_tiles, k_tiles,
+                    ta, tb, M, N, K, beta, **kwargs):
     frac_m = M / m_tiles
     frac_n = N / n_tiles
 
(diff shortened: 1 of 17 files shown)


sw/blas/gemm/src/gemm.h
```diff
@@ -69,55 +76,126 @@ void sc_st_gemm(precision_t prec, uint32_t expand, uint32_t setup_ssr,
...
+                switch (impl) {
+                    case NAIVE:
+                        gemm_fp64_naive(frac_m, n, k, (double*)a + offsetA,
+                                        lda_strided, transa, (double*)b, ldb,
+                                        transb, (double*)c + offsetC,
+                                        ldc_strided, (double)beta);
+                        break;
+                    case NAIVE_UNROLLED:
+                        printf(
+                            "Naive unrolled implementation not supported for "
```
viv-eth, line 88: Move this into validation cfg

sw/blas/gemm/scripts/datagen.py
```diff
@@ -43,15 +44,29 @@ def validate_config(prec, parallelize_m, parallelize_k, m_tiles, n_tiles, k_tile
...
+    assert not ((prec != "FP64") and ((implementation != "BASELINE") or
+                                      (implementation != "NAIVE")) and not tb), \
+        'Optimized SIMD kernels support only transposed B matrix'
     assert not tb or n_tiles == 1, 'Tiling in the N dimension supported only if B is' \
                                    ' not transposed'
     assert not tb or k_tiles == 1, 'Tiling in the K dimension supported only if B is' \
                                    ' not transposed'
-    assert baseline or frac_n >= 8, 'N dimension of tile size must be greater or equal to' \
-                                    ' the unrolling factor (8) when using optimized kernels'
+    assert (implementation != "BASELINE") or (implementation != "NAIVE") or frac_n >= 8, \
```
colluca, line 54: This should be `(implementation == "BASELINE") or (implementation == "NAIVE")`

sw/blas/gemm/src/gemm.h
colluca: I would add that this can be done also for all other unsupported configurations, we can remove them from `gemm.h` to keep it concise, and move it to datagen.

sw/blas/gemm/src/gemm.h
```diff
@@ -11,11 +11,18 @@
...
 // Guard to avoid conflict with DNN header file
 // TODO: move this definition to Snitch math library to solve problem
 #ifndef PRECISION_T
 #define PRECISION_T
 typedef enum { FP64 = 8, FP32 = 4, FP16 = 2, FP8 = 1 } precision_t;
+#endif
+
+#ifndef IMPLEMENTATION_T
+#define IMPLEMENTATION_T
+typedef enum { BASELINE, NAIVE, NAIVE_UNROLLED, OPT, OPT_EX } implementation_t;
```
colluca, line 25: I would use a consistent naming, either `ex_opt` or `opt_ex` everywhere (configuration file names, implementation function signature, datagen script). It used to be `ex_opt` based on the function signature.

colluca (changes requested):
Also would rename all "base" configuration files to "naive"

PR #124, opened 2024-03-22
sw: Add SARIS kernels

paulsc96:
* Adds SARIS stencil kernels to `sw`; see `sw/saris/README.md`.
* Makes `snitch_cluster_wrapper` IO independent of configuration (all ports always available); this was necessary to make sure the testbench fixture works for all configurations.

ToDo:

* [x] Push, reference LLVM15 tree
* [x] Fix Lint
* [x] Fix, check license headers
* [x] Write README
* [x] Fix `putchar` buffer
* [x] Fix TCDM assertion misfires 
  * --> ugly, but benign and independent of SARIS code; separately addressed by #125. 
* Add preprint citation (**except eprint**, lockstep with preprint upload) to:
  * [x] Local README
  * [x] Global README
  * [x] Documentation.
* [x] (Add eprint number) TBD later.


target/snitch_cluster/cfg/sssr.hjson
colluca, line 1: I see quite some differences with the default config, which is not a problem per se but I would suggest we do either of the following:
- Update all non-SSSR related settings in the default (and other) configs with the settings herein
- Rename the config to something like `saris.hjson` just to say that it's specific to this exploration and not a generic SSSR config

.github/workflows/lint.yml
```diff
@@ -129,6 +129,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: DoozyX/clang-format-lint-action@v0.16.2
         with:
+          exclude: './sw/saris'
```
colluca, line 132: Is this strictly necessary?

colluca (changes requested):

target/snitch_cluster/cfg/sssr.hjson
paulsc96: This is a good point. I called it `sssr.json` because this is the config used for all existing (and future) SSSR experiments, but we might aswell call it `occamy.json` because that's what it is (with innovations such as the cluster alias).

Really, the Occamy config should be the default config as the current default is really stupid in some regards (5-stage FPU, slow enforced VM and debugging support, ...) but who knows whose applications will break or tank in performance when you change a default parameter... mine certainly will if you don't use the Occamy config.

In any case, we should probably bite the bullet and switch to Occamy defaults, but we can do that in another MR.

colluca: According to @lucabertaccini, both settings in `default.json` and `sssr.json` produce the same number of pipeline stages in the FPU (so long as FP64 support is enabled), so we can also just keep the settings in `sssr.json`.

Regarding also the other changes, I think we can promote the SSSR settings to the default (and other) configs. I would do it in this PR or another we merge in advance, so we don't need a new config at all.

.github/workflows/lint.yml
paulsc96: Not strictly, no. But currently, the code is formatted for maximum readability of the iterations and point loop. The `clang-format` reformatting really tears it apart and makes it very hard to parse what's going on.

I do agree that this workflow level exclude is ugly; if you know of a way to do it with a pragma or something, we can do that.

target/snitch_cluster/cfg/sssr.hjson
paulsc96: Done. This is now the default config.

I also added some placeholders for the citation; we just need to add the eprint reference in each block.

PR #126, opened 2024-04-17
sw: Add possibility to get ocd putchar on gvsoc

haugoug:
This adds a SELECT_PLATFORM option which can be used to compile the runtime for GVSOC.
This will then select a dedicated putchar which is using OpenOCD semi-hosting. This is also used for exit.
This is used to avoid the polling in dram due to htif modeling, which disturbs dram timing model with extra page misses.
This also fixes an uninitialized access to putchar index which fails on gvsoc since memory is not initialized to zero.

### DIFF
--- a/sw/snRuntime/src/start.c
+++ b/sw/snRuntime/src/start.c
@@ -99,8 +99,12 @@ static inline void snrt_init_libs() { snrt_alloc_init(); }
 #ifdef SNRT_CRT0_EXIT
 static inline void snrt_exit_default(int exit_code) {
     exit_code = snrt_global_all_to_all_reduction(exit_code);
+#ifdef OPENOCD_SEMIHOSTING
+    if (snrt_global_core_idx() == 0) __ocd_semihost(0x18, (long)exit_code);
+#else
     if (snrt_global_core_idx() == 0)
         *(snrt_exit_code_destination()) = (exit_code << 1) | 1;
+#endif
 }
 #ifndef SNRT_CRT0_ALTERNATE_EXIT
 static inline void snrt_exit(int exit_code) { snrt_exit_default(exit_code); }
--- a/target/snitch_cluster/sw/runtime/rtl/src/putchar.c
+++ b/target/snitch_cluster/sw/runtime/rtl/src/putchar.c
@@ -2,6 +2,26 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+#ifdef OPENOCD_SEMIHOSTING
+
+#include <openocd.h>
+
+#define PUTC_BUFFER_LEN 1024
+static int putchar_index[SNRT_CLUSTER_NUM * SNRT_CLUSTER_CORE_NUM] = {0};
+static char putchar_data[SNRT_CLUSTER_NUM * SNRT_CLUSTER_CORE_NUM]
+                        [PUTC_BUFFER_LEN];
+
+// Provide an implementation for putchar.
+void _putchar(char character) {
+    volatile char *buf = putchar_data[snrt_hartid()];
+    buf[putchar_index[snrt_hartid()]++] = character;
+    if (putchar_index[snrt_hartid()] == PUTC_BUFFER_LEN || character == '\n') {
+        __ocd_semihost_write(1, (uint8_t *)buf, putchar_index[snrt_hartid()]);
+        putchar_index[snrt_hartid()] = 0;
+    }
+}
+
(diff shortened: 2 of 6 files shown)


sw/snRuntime/src/start.c
```diff
@@ -99,8 +99,12 @@ static inline void snrt_init_libs() { snrt_alloc_init(); }
 #ifdef SNRT_CRT0_EXIT
 static inline void snrt_exit_default(int exit_code) {
     exit_code = snrt_global_all_to_all_reduction(exit_code);
+#ifdef __GVSOC__
+    if (snrt_global_core_idx() == 0) __semihost(0x18, (long)exit_code);
+#else
     if (snrt_global_core_idx() == 0)
         *(snrt_exit_code_destination()) = (exit_code << 1) | 1;
+#endif
```
colluca, line 107: The files in the `sw/snRuntime` folder are meant to be base sources for multiple different Snitch runtime implementations. Different implementations already exist for [RTL](https://github.com/pulp-platform/snitch_cluster/tree/main/target/snitch_cluster/sw/runtime/rtl) and for [Banshee](https://github.com/pulp-platform/snitch_cluster/tree/main/target/snitch_cluster/sw/runtime/banshee).

I wouldn't implement this change here, but rather create a custom runtime for GVSoC, which defines its own exit function.

target/snitch_cluster/sw/runtime/rtl/src/putchar.c
colluca, line 1: This would also be moved to the GVSoC-specific runtime directory.

target/snitch_cluster/sw/runtime/rtl/src/snitch_cluster_start.c
colluca, line 1: See comments above.

target/snitch_cluster/sw/toolchain.mk
```diff
@@ -42,6 +42,12 @@ RISCV_CFLAGS += -O3
 ifeq ($(DEBUG), ON)
 RISCV_CFLAGS += -g
 endif
+
+ifeq ($(SELECT_PLATFORM), gvsoc)
+RISCV_CFLAGS += -D__GVSOC__
+endif
+
+
```
colluca, line 50: This should also not be necessary anymore. Rather when compiling we have to make sure to reference the sources in the GVSoC-specific runtime. See:

https://github.com/pulp-platform/snitch_cluster/blob/13d1e6b3aaa8e53d5b39858c6ee871b186bf9aa2/target/snitch_cluster/sw.mk#L8-L14

colluca (changes requested):
Hi @haugoug, in general looks good to me.

I made some suggestions to avoid adding complexity to the base runtime sources.

Personally, I think it might be even better to move everything GVSoC-related to a new target subdirectory `target/gvsoc`, and rename `target/snitch_cluster` to `target/rtl` (or `target/sim` as in [Cheshire](https://github.com/pulp-platform/cheshire/tree/main/target/sim)).

If you can provide me with some commands to reproduce the correct behaviour, I could even implement the suggested changes myself.

Let me know what you think :)

target/snitch_cluster/sw/toolchain.mk
haugoug: Actually this is not specific to gvsoc, it's something generic which can be used on any platform, but in particular on boards with openocd. Gvsoc can support any kind of printf, currently both the tohost/fromhost support of the runtime and also openocd.
Probably the name is misleading, I can rename it to SELECT_IO which would by default be htif, and could be overwritten to ocd ?

target/snitch_cluster/sw/runtime/rtl/src/snitch_cluster_start.c
haugoug: See my reply below.

target/snitch_cluster/sw/toolchain.mk
colluca: I see, thanks for the clarification. In that case I agree, and I would do two things, if you think it would make sense:
- move the OpenOCD related functions (e.g. `__semihost`) to an `openocd.h` file which would be included by a Snitch runtime implementation only if semihosting is supported on that target
- rename the `__GVSOC__` flag to something like `OPENOCD_SEMIHOSTING`

haugoug: Yes it makes sense, I've pushed the modifications, let me know.

colluca (approved):
LGTM! Will merge as soon as CI passes.

Thanks for the contribution @haugoug :)

PR #133, opened 2024-05-01
sw: Add missing `extern` declarations

zero9178:
Without these `extern` declarations, the final object file will not contain functions marked `inline` as LLVM discards them prior to code generation.

This has so far not been problematic as these methods usually get inlined. Downstream projects as well as consumers of the `api/` headers may, however, be using different compiler flags or not have inline definitions available within the same translation unit. This would then cause an "undefined reference" linker error in any such project.

### DIFF
--- a/sw/snRuntime/src/alloc.c
+++ b/sw/snRuntime/src/alloc.c
@@ -4,7 +4,7 @@
 
 snrt_allocator_t l3_allocator;
 
-extern volatile uint32_t* snrt_zero_memory_ptr();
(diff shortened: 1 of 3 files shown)


viv-eth (approved):
LGTM! Thank you.

colluca:
Thanks for your contribution @zero9178!

I'm closing this PR in favour of https://github.com/pulp-platform/snitch_cluster/pull/136 where I included your changes.

PR #134, opened 2024-05-01
sw: Fix warning in `kmp.c`

zero9178:
Compiling the file would previously emit 
```
warning: missing terminating '"' character [-Winvalid-pp-token]
```
due to what I assume is a mistake introduced through some auto formatter. Multi line strings are actually illegal in C and Clang and GCC are simply kind enough to only emit a warning. The C does however implicitly concatenate two string literals appearing next to each other producing the intended result.


viv-eth (approved):
LGTM! Thank you for the contribution @zero9178

colluca:
Thanks for your contribution @zero9178!

I'm closing this PR in favour of https://github.com/pulp-platform/snitch_cluster/pull/135 where I included your changes.

PR #138, opened 2024-05-10
target: Allow verilator build with musl libc

zero9178:
Building the snitch verilator model on distros using musl libc is currently not possible due to some glibc specific assumption and a musl libc peculiarly. More specifically, code was using `#include <printf.h>` instead of `#include <stdio.h>` which is a glibc specific header. Lastly, `PAGE_SIZE` is a macro provided by Linux headers that expands to a number. musl libc unfortunately indirectly includes it via `#include <limit.h>`. Renaming occurrences of `PAGE_SIZE` fixes the name clash.

### DIFF
--- a/target/common/test/tb_bin.cc
+++ b/target/common/test/tb_bin.cc
@@ -2,7 +2,7 @@
 // Solderpad Hardware License, Version 0.51, see LICENSE for details.
 // SPDX-License-Identifier: SHL-0.51
 
-#include <printf.h>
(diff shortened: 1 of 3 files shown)


colluca:
Once again, thanks for the contribution @zero9178 :)

It LGTM! We are currently facing issues with the internal Gitlab CI. I will fix it ASAP and then proceed to merge this.

colluca (approved):

PR #142, opened 2024-05-16
Gvsoc ci

haugoug:
[@colluca edit] Extends the Snitch cluster repo to enable Snitch cluster testing in the GVSoC repo's CI.


util/sim/Simulator.py
```diff
@@ -96,6 +96,23 @@ def get_simulation(self, test):
...
+    [Gvsoc simulations][Simulation.GvsocSimulation].
+    """
+
+    def __init__(self, binary):
+        """Constructor for the GvsocSimulator class.
+
+        Arguments:
+            binary: The Gvsoc simulation binary.
+        """
+        super().__init__(binary, name='gvsoc', simulation_cls=GvsocSimulation)
```
colluca, line 113: Why does it inherit from the `RTLSimulator` class?

target/snitch_cluster/sw/apps/common.mk
```diff
@@ -16,6 +16,11 @@ include $(MK_DIR)/../toolchain.mk
 # Fixed paths in repository tree
 ROOT     := $(abspath $(MK_DIR)/../../../..)
 SNRT_DIR := $(ROOT)/sw/snRuntime
+
+ifeq ($(BIST), ON)
+RISCV_CFLAGS += -DBIST
+endif
+
```
colluca, line 23: Is there an advantage in passing `BIST=ON` on the command-line rather than `RISCV_CFLAGS=-DBIST`? Also, where is this used?

colluca (changes requested):
Thanks for the contribution @haugoug!

I have a few major comments:
* To add the GVSoC tests to the CI the workflow file needs to be adapted as well, see https://github.com/pulp-platform/snitch_cluster/blob/e90dceb6a6c9bdb6bb5d1502b62b920242be9fa8/.github/workflows/ci.yml#L47
* For all tests which make use of the IPC verification interface (i.e. those which have no BIST verification), just running the binaries with GVSoC doesn't really check anything. Perhaps we want to extend the IPC verification interface to work also with GVSoC, otherwise we just need to make clear that the GVSoC simulation will not really verify those tests.

Minor comments follow on the code.

target/snitch_cluster/sw/apps/common.mk
haugoug: No Indeed I will remove it.

util/sim/Simulator.py
haugoug: I have changed it to inherit from Simulator class

PR #146, opened 2024-05-30
ci: Fix verilator environment variables for IIS machines

fischeti:


viv-eth (approved):

Xeratec:
As both other reviewers are on holidays and this fix/hack is urgent, I will merge it and ask for forgiveness.

PR #137, opened 2024-05-08
sw: Restructuring of BLAS and DNN kernels

viv-eth:
This PR will introduce substantial changes to the SW library in the BLAS and DNN kernels. 

## GEMM

This PR comprises the following changes in the BLAS library: 

1. We moved to a generic `impl` function that will fetch the correct kernel from the `config.json` file. All of this is wrapped inside a `gemm_args_t` struct that holds all kernel-related data. The arguments in the struct are loaded into the cluster's TCDM to avoid costly DRAM accesses when calling the function. The roadmap is to adapt all kernels in the `sw` library to this format to improve performance. 
2. We added the option to run multiple tests in parallel. 
3. We aligned the tests to work with the new GEMM layout and added a naive FP16 kernel for verification. 
4. We added more assertions to the `datagen` script of unsupported or invalid configurations with the relevant error messages.
5. We adjusted other kernels, such as `fused_concat_linear` to the new layout.

## FlashAttention-2 

1. We added the transposition layer that is used inside the kernel.
2. We added the FP16 and FP8 MiniFloat implementations. 
3. We added multiple configurations for testing. 
4. We aligned the kernel to work with the new GEMM layout.
5. Several fixes to the kernel were integrated and casts from lower to higher precision and vice versa.

## Treewide

This PR also includes some treewide changes.

1. Remove all math library related sources, which are no longer needed after https://github.com/pulp-platform/snitch_cluster/pull/90.

2. Remove all uses of the `$PYTHON` environment variable, and instead consistently rely on shebang directives to pick up the active virtual environment's `python` executable.

PR #147, opened 2024-06-05
sw: Multi-cluster runtime improvements

fischeti:
This PR aims to improve the snitch runtime when when the snitch cluster is used as a dependency in a multi-cluster system.

### Changes

* Autogenerate `snitch_cluster_defs.h` with the number of clusters
* Getting rid of `nr_quadrants` in configuration, which is obsolete in my opinion and was never actually used anywhere.


.gitlab-ci.yml
```diff
@@ -35,7 +35,9 @@ snitch-cluster-sw:
   artifacts:
     paths:
       - sw/math/include/bits/alltypes.h
+      - target/snitch_cluster/sw/runtime/common/*.h
       - target/snitch_cluster/sw/**/build/*.elf
+      - target/snitch_cluster/sw/runtime/rtl/build/libsnRuntime.a
```
colluca, line 40: Why is this needed? Is there any dependent CI stage job which uses these?

docs/schema/snitch_cluster.schema.json
```diff
@@ -14,6 +14,11 @@
             "description": "Optional name for the generated wrapper.",
             "default": "snitch_cluster"
         },
+        "nr_clusters": {
+            "type": "number",
+            "description": "Number of clusters. Only relevant for the SW runtime.",
+            "default": 1
+        },
```
colluca, line 21: Why is this added as a property of the Snitch cluster?

colluca (changes requested):
Since `snitch_cluster_defs.h` is also auto-generated now, perhaps we want to just merge it with `snitch_cluster_cfg.h`. I believe the previous idea was to just separate the auto-generated defines to a single place, and everything else derived from it. But I think it's not too relevant and we can just merge them to possibly simplify the codebase.

Further minor comments follow.

.gitlab-ci.yml
fischeti: There was an issue, but I don't remember anymore why it failed. I will check again

target/common/test/tb_lib.hh
```diff
@@ -126,7 +126,6 @@ struct BootData {
     uint64_t global_mem_start;
     uint64_t global_mem_end;
     uint32_t cluster_count;
-    uint32_t s1_quadrant_count;
```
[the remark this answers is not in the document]
fischeti: No, I haven't checked that. But doesn't Occamy use a different configuration file resp. schema anyway?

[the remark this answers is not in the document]
colluca: Yeah I think we can remove this Occamy-specific parameter from this repo, and later align Occamy if needed.

.gitlab-ci.yml
fischeti: I remember the issue again. The [`run.py`](https://github.com/pulp-platform/snitch_cluster/blob/17aa42a7a002c633e91b5948a94b71404d1bd5d9/target/snitch_cluster/sw/apps/blas/gemm/test/run.py#L40) script builds the software including runtime again, but it does not generate the header files again (see [here](https://iis-git.ee.ethz.ch/github-mirror/snitch_cluster/-/jobs/1350489))

colluca: I see, thanks. I will look into this further, and otherwise we can merge it as is.

fischeti: I reverted the commit to check it, but let me know if I should rerevert it to merge it😉

colluca: Thanks Tim :) I think we will be able to fix this in https://github.com/pulp-platform/snitch_cluster/pull/137 where we are anyways slightly generalizing the `run.py` script. So no need to revert it now, I think we can get this PR to pass as is once we rebase on the other one.

PR #148, opened 2024-06-05
target: Add `VLT_NUM_THREADS` to configure threads used by verilator

zero9178:
Using more threads in the simulation may lead to a speedup in simulation time, especially when the program running in the simulator makes use of multiple cores. This PR therefore adds the `VLT_NUM_THREADS` option to `make` to allow users to set the thread number used in the simulator. For backwards compatibility, and since the number of threads is environment and code dependent, the default value of 1 was kept.

### DIFF
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -44,6 +44,7 @@ EVENTVIS_PY      ?= $(UTIL_DIR)/trace/eventvis.py
 # a two-liner with the OS on the first line, hence the tail -n1
 VERILATOR_ROOT ?= $(dir $(shell $(VERILATOR_SEPP) which verilator | tail -n1))..
 VLT_ROOT       ?= ${VERILATOR_ROOT}
+VLT_NUM_THREADS ?= 1
 
 MATCH_END := '/+incdir+/ s/$$/\/*\/*/'
 MATCH_BGN := 's/+incdir+//g'
@@ -86,7 +87,13 @@ VLT_FLAGS    += -Wno-UNSIGNED
 VLT_FLAGS    += -Wno-UNOPTFLAT
 VLT_FLAGS    += -Wno-fatal
 VLT_FLAGS    += --unroll-count 1024
+ifneq ($(VLT_NUM_THREADS), 1)
+	VLT_FLAGS	 += --threads $(VLT_NUM_THREADS)
+endif
 VLT_CFLAGS   += -std=c++14 -pthread
+ifneq ($(VLT_NUM_THREADS), 1)
+	VLT_CFLAGS	 += -DVL_THREADED=1
+endif
 VLT_CFLAGS   +=-I ${VLT_BUILDDIR} -I $(VLT_ROOT)/include -I $(VLT_ROOT)/include/vltstd -I $(VLT_FESVR)/include -I $(TB_DIR) -I ${MKFILE_DIR}/test
 
 ANNOTATE_FLAGS      ?= -q --keep-time --addr2line=$(ADDR2LINE)
--- a/target/common/common.mk
+++ b/target/common/common.mk
@@ -87,13 +87,11 @@ VLT_FLAGS    += -Wno-UNSIGNED
 VLT_FLAGS    += -Wno-UNOPTFLAT
 VLT_FLAGS    += -Wno-fatal
 VLT_FLAGS    += --unroll-count 1024
-ifneq ($(VLT_NUM_THREADS), 1)
-	VLT_FLAGS	 += --threads $(VLT_NUM_THREADS)
-endif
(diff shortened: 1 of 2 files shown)


target/common/common.mk
```diff
@@ -86,7 +87,13 @@ VLT_FLAGS    += -Wno-UNSIGNED
 VLT_FLAGS    += -Wno-UNOPTFLAT
 VLT_FLAGS    += -Wno-fatal
 VLT_FLAGS    += --unroll-count 1024
+ifneq ($(VLT_NUM_THREADS), 1)
+	VLT_FLAGS	 += --threads $(VLT_NUM_THREADS)
+endif
```
colluca, line 92: I believe we can drop the `if` statement, as [Verilator should by default behave as if `--threads 1` is passed](https://github.com/pulp-platform/snitch_cluster/assets/17752123/7863d1e1-479b-4dfb-a83a-97c2bb6e5b28).

target/common/common.mk
```diff
@@ -86,7 +87,13 @@ VLT_FLAGS    += -Wno-UNSIGNED
...
 VLT_CFLAGS   += -std=c++14 -pthread
+ifneq ($(VLT_NUM_THREADS), 1)
+	VLT_CFLAGS	 += -DVL_THREADED=1
+endif
```
colluca, line 96: Couldn't find much information on this, but perhaps we can also always add this flag, and remove the if statement. I guess it shouldn't break the single-threaded case.

target/snitch_cluster/Makefile
```diff
@@ -103,6 +103,9 @@ VLT_COBJ += $(VLT_BUILDDIR)/tb/tb_bin.o
 VLT_COBJ += $(VLT_BUILDDIR)/vlt/verilated.o
 VLT_COBJ += $(VLT_BUILDDIR)/vlt/verilated_dpi.o
 VLT_COBJ += $(VLT_BUILDDIR)/vlt/verilated_vcd_c.o
+ifneq ($(VLT_NUM_THREADS), 1)
+	VLT_COBJ += $(VLT_BUILDDIR)/vlt/verilated_threads.o
+endif
```
colluca, line 108: Same as for comments above.

colluca (changes requested):
Thanks for the valuable contribution @zero9178!

I have just a few comments to simplify the change further. Let me know if you think it could work.

target/common/common.mk
zero9178: I believe this is true for Verilator 5 but not the verilator 4 that the repo currently supports until https://github.com/pulp-platform/snitch_cluster/pull/76 lands.
That said I think you're right and we can drop this as this is almost certainly a premature optimization and I doubt single-threaded usage will see a measureable difference.

target/common/common.mk
zero9178: Information is sadly very sparse on the flag and I only found it referenced through GitHub issues such as https://github.com/verilator/verilator/issues/3668#issuecomment-1288989145
I had to add it as I was getting compilation errors as verilator requires it to be defined when using multiple threads. 
I added a comment to clarify.

zero9178:
I am seeing issues with the verilator build where it seemingly gets stuck/deadlocks for non-obvious reasons. While the PR itself is okay its probably not a good idea to actually use any value but 1 for the time being.

colluca:
> I am seeing issues with the simulation where it seemingly gets stuck/deadlocks for non-obvious reasons. While the PR itself is okay its probably not a good idea to actually use any value but 1 for the time being.

I believe @and-ivanov observed something like this in the past. Tagging him in case he might have further comments.

colluca:
Dear @zero9178,

I opened a new PR upgrading the supported Verilator version to 5.006 (https://github.com/pulp-platform/snitch_cluster/pull/158). I tentatively added your multi-threading contributions from this PR. If you would like to try it, perhaps it fixes the issues you mentioned above. If this is the case we can close this PR in favour of https://github.com/pulp-platform/snitch_cluster/pull/158 and merge it.

Best,
Luca

zero9178:
> Dear @zero9178,
> 
> I opened a new PR upgrading the supported Verilator version to 5.006 (#158). I tentatively added your multi-threading contributions from this PR. If you would like to try it, perhaps it fixes the issues you mentioned above. If this is the case we can close this PR in favour of #158 and merge it.
> 
> Best, Luca

That's amazing! Let me run the verilator model from your PR through our stack and I'll get back to you whether it works (I'll respond on that PR). If yes then I think we can close this. If the multithreading part does not work then I'd either leave this open and/or make this an issue.

colluca:
Finally merged in https://github.com/pulp-platform/snitch_cluster/pull/158.

PR #151, opened 2024-06-11
sw: Add DOTP benchmark

mp-17:
And after all the noise around it, I share the implementation of DOTP that I am using.
Feel free to check it, fix it if needed, and add further optimizations. 

Let me know if there are any mistakes.

NOTE: as of now, the kernel only supports a single cluster.


colluca:
Thanks @mp-17! I can take it on from here :)

sw/blas/dotp/scripts/datagen.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def emit_header(self, **kwargs):
+        header = [super().emit_header()]
+
+        n = kwargs['n']
+        x = np.random.uniform(self.MIN, self.MAX, n)
+        y = np.random.uniform(self.MIN, self.MAX, n)
+        g = self.golden_model(x, y)
+
+        assert (n % 8) == 0, "n must be an integer multiple of the number of cores"
```
fischeti, line 34: For the unrolled kernel, each fraction of the vector also needs to be a multiple of 4 right?

sw/blas/dotp/src/main.c
```diff
@@ -0,0 +1,110 @@
...
+    snrt_cluster_hw_barrier();
+
+    if (!snrt_cluster_core_idx()) {
+      unsigned int runtime = end_cycle - start_cycle;
+      double performance   = (double) (2 * n - 1) / runtime;
+      double util          = 100 * (performance / (2 * snrt_cluster_compute_core_num()));
+
+      printf("Core %d execution time: %u cycles\nPerformance: %f DP-FLOP/Cycle\nUtilization: %f%%\n",
+        snrt_cluster_core_idx(), runtime, performance, util);
+    }
```
fischeti, line 82: Usually we extract performance metrics from the traces. But we report the utilization based on instructions and not FLOPs there.

sw/blas/dotp/scripts/datagen.py
colluca: Good catch! I updated the script.

fischeti (approved):
LGTM. What would be nice is `fp8`, `fp16`, `fp32` support, but we can leave this for later.

colluca (approved):

PR #153, opened 2024-06-28
sw: Fix `snrt_dma_wait` implementation

zero9178:
The previous implementation of `snrt_dma_wait` was incorrect as it had an infinite loop in the case that `tid` was the last completed transaction. Looking at the documentation, the snippet in custom_instructions.md shows the correct way of implementing the transaction check (get last completed tid and loop if it's less than the one we are waiting for) except it had its register operands swapped. This PR uses that implementation instead, fixes the small bug in the documentation and changes one of the waits in `dma_simple.c` to make sure we don't regress either.

### DIFF
--- a/docs/rm/custom_instructions.md
+++ b/docs/rm/custom_instructions.md
@@ -104,7 +104,7 @@ The DMSTATI instruction can be used to implement a blocking wait for the complet
 
         dmcpyi a0, ...
     1:  dmstati t0, 0
-        bltu a0, t0, 1b
+        bltu t0, a0, 1b
 
 Similarly, waiting for the completion of *all* DMA transfers:
 
--- a/sw/snRuntime/src/dma.h
+++ b/sw/snRuntime/src/dma.h
@@ -151,8 +151,7 @@ inline void snrt_dma_wait(snrt_dma_txid_t tid) {
                (    0b000 << 12) | \
                (      (5) <<  7) | \
                (0b0101011 <<  0)   \n"
-        "sub t0, t0, %0 \n"
-        "blez t0, 1b \n" ::"r"(tid)
+        "bltu t0, %0, 1b \n" ::"r"(tid)
         : "t0");
 }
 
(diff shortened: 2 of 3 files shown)


fischeti (approved):
LGTM. Thanks for the fix!

zero9178:
> LGTM. Thanks for the fix!

Thank you! Would you mind merging this for me as well? I do not have commit writes :)

fischeti:
> > LGTM. Thanks for the fix!
> 
> Thank you! Would you mind merging this for me as well? I do not have commit writes :)

Me neither unfortunately😉 @paulsc96 and @colluca are maintainers of the repository

colluca (changes requested):
Thanks for the contribution @zero9178 :)

Looking at the PR, it looks like this breaks the way zero-sized transfers are handled (return a transaction ID of -1).

I'm adding a test for these transfers and correcting the condition. Will merge as soon as I get this fixed :)

zero9178:
> Thanks for the contribution @zero9178 :)
> 
> Looking at the PR, it looks like this breaks the way zero-sized transfers are handled (return a transaction ID of -1).
> 
> I'm adding a test for these transfers and correcting the condition. Will merge as soon as I get this fixed :)

Thanks good catch I did not consider these. Could these use 0 instead or does the DMA engine return 0 for the very first transaction? Otherwise I'd just change the `bltu` to `blt`

colluca:
> Thanks good catch I did not consider these. Could these use 0 instead or does the DMA engine return 0 for the very first transaction? Otherwise I'd just change the `bltu` to `blt`

Zero seems to work. But actually the new DMA should support zero-sized transfers, so removing the if condition on the size altogether should work. There seems to be a bug however which I'm currently investigating.

colluca:
A fix to the DMA bug is provided in https://github.com/pulp-platform/iDMA/pull/50. Till that is merged and a new release of the iDMA is created, we can merge this PR with the zero ID as suggested.

colluca (approved):

PR #155, opened 2024-07-03
hw: Improve performance counters for bring-up

fischeti:
This PR adds improvements to the performance counters for bring-up, and restructures the performance counter implementation:

### Changed

* After reset, performance counters immediately start tracking, which allows checking from outside if the cores, icache, etc., are alive. The metrics that are currently tracked are: 
https://github.com/pulp-platform/snitch_cluster/blob/333ee0d619268a59f987e9466f281dba075a7bd6/hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv#L103-L111

* The configuration registers are refactored for a more sensible encoding of the metrics that can be tracked. Further, the configuration of the metric/hart and the enable are now separate registers, which should make it a bit faster resp. more accurate to enable/disable performance counters. This means the interface for the performance counter API changed slightly, but makes more sense now.

* The previous `perf_cnt` test was adapted to the new performance counter API interface, and now actually checks for error. Previously, it served mostly as an example on how to set up the performance counters and always returned zero.

### TODO
- [x] Enable tracking of multi-channel DMA as discussed in https://github.com/pulp-platform/snitch_cluster/pull/144
- [x] ~Revert container references in CI to `main` when deemed ready to merge~
- [ ] Credit initial authors once it is merged:
```
Co-authored-by: Milos Hirsl <mhirsl@student.ethz.ch>
```


sw/tests/perf_cnt.c
```diff
@@ -5,110 +5,134 @@
...
+            snrt_reset_perf_counter(i);
 
-        // Start performance counter
-        snrt_start_perf_counter(SNRT_PERF_CNT0, SNRT_PERF_CNT_CYCLES, 0);
+            // Check that the performance counter is reset
+            errors += (snrt_get_perf_counter(i) != 0);
+        }
+    }
+
+    // Test 3: Check that the performance counters can configured and started
```
colluca, line 33: can be*

colluca (changes requested):
Thanks for the contribution Tim! Mostly LGTM, just a few comments following.

hw/snitch_cluster/src/snitch_cluster_peripheral/Makefile
```diff
@@ -4,15 +4,16 @@
...
+REGTOOL=$(shell $(BENDER) path register_interface)/vendor/lowrisc_opentitan/util/regtool.py
 PYTHON3=$(shell which python)
 CLANG_FORMAT=$(shell which clang-format-12.0.1)
 
 SCHEMA=snitch_cluster_peripheral_reg.hjson
 
 PKG=snitch_cluster_peripheral_reg_pkg.sv
 TOP=snitch_cluster_peripheral_reg_top.sv
-CDEF=../../../../../sw/snRuntime/include/snitch_cluster_peripheral.h
+CDEF=../../../../target/snitch_cluster/sw/runtime/common/snitch_cluster_peripheral.h
```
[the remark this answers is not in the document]
fischeti: True, I also thought that those are kind of duplicate. I actually moved the whole generation to the snitch cluster Makefile now, let me know if you like it or not.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.hjson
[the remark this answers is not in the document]
fischeti: I will try to remove it. I am not sure if there is an issue with banshee, which still uses it afaik.

sw/snRuntime/src/perf_cnt.h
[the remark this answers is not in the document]
fischeti: I added the `perf_cnt.c` file

sw/snRuntime/src/perf_cnt.h
```diff
@@ -2,100 +2,55 @@
...
 } perf_reg32_t;
 
 typedef struct {
-    volatile perf_reg32_t enable[SNRT_PERF_N_CNT];
-    volatile perf_reg32_t hart_select[SNRT_PERF_N_CNT];
-    volatile perf_reg32_t perf_counter[SNRT_PERF_N_CNT];
+    volatile perf_reg32_t enable[SNRT_NUM_PERF_CNTS];
+    volatile perf_reg32_t select[SNRT_NUM_PERF_CNTS];
+    volatile perf_reg32_t perf_counter[SNRT_NUM_PERF_CNTS];
 } perf_regs_t;
```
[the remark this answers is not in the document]
colluca: Indeed it made sense (and I think it's better) to have a union with a struct internally. But this is only needed on the right-hand side of the assignment in `snrt_cfg_perf_counter()`, where you now use the `<<` and `|` ops instead.

If we really need the `perf_reg32_t` type, can't we just use a `typedef` instead of a union? Accessing `.value` everywhere seems unnecessary to me. But I also still don't understand why we need this custom type at all rather than just a common `uint32_t`.

target/snitch_cluster/Makefile
```diff
@@ -47,6 +47,7 @@ CLUSTER_GEN_SRC ?= $(wildcard $(ROOT)/util/clustergen/*.py)
 
 BIN_DIR           ?= bin
 GENERATED_DIR     ?= $(MKFILE_DIR)generated
+PERIPH_DIR        ?= $(ROOT)/hw/snitch_cluster/src/snitch_cluster_peripheral
```
colluca, line 50: I think we can stop tracking the generated `snitch_cluster_peripherals` files and move them to `GENERATED_DIR` as well.

target/snitch_cluster/Makefile
```diff
@@ -164,6 +165,11 @@ $(GENERATED_DIR)/memories.json: ${CFG} ${CLUSTER_GEN_PREREQ} | $(GENERATED_DIR)
 $(GENERATED_DIR)/bootdata.cc: ${CFG} ${CLUSTER_GEN_PREREQ} | $(GENERATED_DIR)
 	$(CLUSTER_GEN) -c $< -o $(GENERATED_DIR) --bootdata
 
+# REGGEN regfile
+$(PERIPH_DIR)/snitch_cluster_peripheral_reg_top.sv: $(PERIPH_DIR)/snitch_cluster_peripheral_reg.hjson
+	@echo "[REGGEN] Generating $@"
+	$(REGGEN) -r -t $(PERIPH_DIR) $<
+
```
colluca, line 172: We should probably either have a PHONY `rtl` target, which is invoked by `all` and generates this file, together with the cluster wrapper and so on, or we should add them as a dependency of the relevant targets, e.g. `bin/snitch_cluster.vsim` (possibly grouped in an `RTL_SOURCES` variable). Best option would be to have both.

Also `snitch_cluster_peripheral_reg_top.sv` is not the only file generated by this command right? So the others should be listed next to it, before the `:`, as well.

target/snitch_cluster/Makefile
fischeti: Yeah, why not. I will change it

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.hjson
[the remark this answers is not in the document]
fischeti: banshee tests are all passing, so I would say that worked

sw/snRuntime/src/perf_cnt.h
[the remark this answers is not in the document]
fischeti: The custom `perf_reg32_t` is useful here because all the configuration registers are 8 byte aligned, eventhough only the lower 32 bits are needed to configure the performance counter (the actual performance counter are 48-bit). With the `perf_reg32_t` arrays you can directly index with the performance counter index passed to the functions, which is kind of neat imo. I also tried with a standard `typedef` but then it only aligns the start of the arrays by 8-byte but not all the elements.

target/snitch_cluster/Makefile
fischeti: I moved them to the `generated` directory now

target/snitch_cluster/Makefile
fischeti: I added a `rtl` target. I think the `RTL_SOURCES` is already done now anyway, or not?

sw/snRuntime/src/perf_cnt.h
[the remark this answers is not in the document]
colluca: Clear thanks :)

target/common/common.mk
```diff
@@ -47,14 +47,10 @@ VLT_ROOT        ?= ${VERILATOR_ROOT}
 VLT_JOBS        ?= $(shell nproc)
 VLT_NUM_THREADS ?= 1
 
-MATCH_END := '/+incdir+/ s/$$/\/*\/*/'
-MATCH_BGN := 's/+incdir+//g'
-SED_SRCS  := sed -e ${MATCH_END} -e ${MATCH_BGN}
-
```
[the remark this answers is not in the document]
fischeti: ah no, I think I missunderstood the sed command. They are still needed for the include files. I will revert it

target/snitch_cluster/Makefile
```diff
@@ -149,6 +150,9 @@ include $(ROOT)/target/snitch_cluster/sw.mk
 # RTL #
 #######
 
+.PHONY: rtl
+rtl: $(GENERATED_DIR)/snitch_cluster_wrapper.sv $(GENERATED_DIR)/snitch_cluster_peripheral_reg_pkg.sv
+
```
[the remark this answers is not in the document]
fischeti: will add it 👍

target/snitch_cluster/Makefile
fischeti: The command generates both `reg_pkg` and `top_pkg` at the same time, so when you add two targets, it will do it twice. Instead I now added one of them as a prerequisite of the other to avoid this.

fischeti: But something still does not work with the prerequisites somehow in the CI. I will check what is wrong

colluca: Yeah, sorry I added the explanation before, but eliminated it by mistake. Give me 30mins and I'll write it again

target/common/common.mk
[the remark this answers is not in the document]
fischeti: Actually, it is indeed unnecessary. the `flist` bender command does not print out any `+incdir+`, only the `flist-plus` does. So either we can remove the `sed` or we should actually use `flist-plus`. But

target/snitch_cluster/Makefile
colluca: The `rtl` target is convenient because you can separately choose to build only the RTL sources. However, it has to be separately invoked before the simulation models can be built, or this would fail cause the files aren't available. This means we need to add it to the CI step, to the tutorial and also to our flow every time those files have to be updated.

In my opinion, it is just much better to have this done transparently by the build system behind the scenes, properly encoding the dependencies in the Makefiles. This would be done by adding the `RTL_SOURCES` as a prerequisite of all targets which depend on them (i.e. those whose rules would fail if the files were missing).

target/common/common.mk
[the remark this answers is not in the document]
colluca: Whatever you think is cleaner :)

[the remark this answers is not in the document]
fischeti: I will remove the `sed` then

Bender.yml
```diff
@@ -149,13 +149,17 @@ sources:
       - hw/snitch_ssr/test/tb_simple_ssr.sv
       - hw/snitch_ssr/test/tb_simple_ssr_streamer.sv
 
+  # cluster peripherals
+  - target: snitch_cluster
+    files:
+      - target/snitch_cluster/generated/snitch_cluster_peripheral_reg_pkg.sv
+      - target/snitch_cluster/generated/snitch_cluster_peripheral_reg_top.sv
+      - hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv
+
```
colluca, line 158: There's actually already a `snitch_cluster` target, see below. But I'm not sure we can use the same target for both. Consider e.g. Occamy, where we actually don't want to include the `snitch_cluster` target with the `snitch_cluster_wrapper` file, cause it generates its own cluster wrapper.

Is this the same also in the FlooNoC-based system?

target/snitch_cluster/Makefile
colluca: Ah yes, you're right about that.

I see the fix, and it's quite tricky. Do you understand why it works after moving this up `COMMON_BENDER_FLAGS += -t snitch_cluster`?

I would've expected that the expansion in the right-hand side of e.g. the `VSIM_SOURCES` assignment would be deferred, and thus expanded after the addition to `COMMON_BENDER_FLAGS` had already been read, so the order wouldn't matter https://www.gnu.org/software/make/manual/html_node/Reading-Makefiles.html.

iis-setup.sh
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Define environment variables
-export BENDER=bender-0.27.1
+export BENDER=bender-0.28.1
 export CC=gcc-9.2.0
```
colluca, line 8: This most likely has to be updated also in the `Dockerfile`.

Bender.yml
colluca: This is actually the only reason why we didn't add the `snitch_cluster` target directly to the `COMMON_BENDER_FLAGS` in the `common.mk` file. As this would be used by derived systems e.g. Occamy, but those don't necessarily include `snitch_cluster_wrapper.sv`.

fischeti: I see, in that case I will not specify a target for the peripheral registers. I thought it might make sense to "bundle" the auto-generated sources with the `snitch_cluster` target

fischeti: The problem then might be that derived systems first need to generate the regfile sources if they are not being tracked. Maybe it would make more sense to keep them in the `snitch_cluster_peripheral` directory and track them. What do you think?

colluca: Perhaps it would make sense, maybe for derived systems, but let's see then when we have a need and use-case for this. Just we would have to have two distinct targets.

target/snitch_cluster/Makefile
fischeti: Actually no, according to the manual I would also expect that it does not make a difference

Bender.yml
colluca: Yes indeed, I don't know what the best approach is in this regard.

1) Probably the simplest is to just generate them in `PERIPH_DIR` and track them (but ideally the CI would then check that the tracked files are always up-to-date with the generated ones).

2) The best way would probably be to do like the DMA, where the files are generated and distributed with releases, but this is more complex to set up, and I guess it relies on versioning.

3) A compromise solution could be to not track them and let derived systems generate them. Anyways, they would ideally reuse the Make snippets provided in `common`. We could think of just moving the peripherals generation there.

Maybe we can just go with the first for now, to start merging something. But I'm also good with the third, as you prefer.

target/snitch_cluster/Makefile
fischeti: I will resolve this, since I reverted the commit. the `snitch_cluster` target is not needed anymore for the peripherals.

Bender.yml
fischeti: I went with the first option for now, and the generated regfiles are tracked in the `PERIPH_DIR` folder. I also added a CI diff porcelain check to make sure the sources are always up to date.

iis-setup.sh
fischeti: I will update bender in a separate PR. I reverted all the related changes, since they are not strictly needed for this PR.

colluca (approved):
LGTM :)

PR #165, opened 2024-07-15
sw: Fixed wrong return value when using ocd semihosting

haugoug:


haugoug:
#142  is not ready yet, I would prefer to quickly merge this one, Sergio needs this fix.

colluca (approved):

Issue #13, opened 2021-08-03
snRuntime: snrt_putchar is not thread safe


huettern:
Communication with `tohost`/`fromhost` is not thread safe and should be guarded with a mutex

colluca:
Fixed in https://github.com/pulp-platform/snitch_cluster/pull/166.

PR #166, opened 2024-07-18
sw: Runtime Fixes

Xeratec:
This PR introduces two fixes to the Snitch runtime that are required for Deeploy.

The first issue is related to the location of the `snrt_l1_allocator()` which is calculated via the `cls()` function. This function uses the `_cls_ptr` which is placed at the beginning of the `.cbss` section. However, if the `.cbss` section is empty, this location is invalid.

The second fix statically allocated the buffers for multicore compatible `printf()` function. Without this fix, the buffers overlap with the L3 allocator, causing `printf()` call to overwrite the allocation structure.

## Fixed
- Ensure the `.cbss` section is not optimized away.
-  Statically allocate `printf()` buffers

### DIFF
--- a/sw/snRuntime/base.ld
+++ b/sw/snRuntime/base.ld
@@ -66,7 +66,7 @@ SECTIONS
   .cbss :
   {
     __cbss_start = .;
-    *(.cbss .cbss.*)
+    KEEP(*(.cbss .cbss.*))
     __cbss_end = .;
   } >L3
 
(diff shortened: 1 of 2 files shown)


fischeti (approved):
LGTM!

colluca (changes requested):
This LGTM, in terms of equivalence. However, the `putchar` function was already broken to start with, see https://github.com/pulp-platform/snitch_cluster/issues/13.

I will push a commit where I fixed this on top and we can merge this together.

colluca (approved):

PR #171, opened 2024-07-23
sw: Switch to non-recursive Make build system

colluca:
This PR implements a shift from the recursive Make paradigm to a non-recursive Make build system for Snitch software. The advantages of a non-recursive implementation are well-known.

The shift was triggered by a limitation which emerged in https://github.com/pulp-platform/snitch_cluster/pull/147. As the dependency between an app and the snRuntime is not represented in the Makefiles (they are distinct subdirectories which are built recursively), the runtime has to be manually built before an application can be built. This dependency can be trivially encoded in a non-recursive implementation.

This PR preserves all the functionalities of the old build system, though the exact commands and API may have changed, namely:

1. PHONY targets to build and clean all software artifacts (i.e. `make sw`, `make clean-sw`)
2. PHONY targets to build and clean each individual software subdirectory (as defined in the old recursive implementation), i.e. each app, the runtime and all tests (e.g. `make gemm`, `make runtime`, `make tests`)
3. possibility to build each app with a non-default data configuration, defined by an environment variable
4. possibility to build each app in a non-default build directory, defined by an environment variable

The latter three, which are needed to implement extensive regression tests, are used e.g. in `target/snitch_cluster/sw/apps/blas/gemm/test`.

It further improves on code reuse, sharing common parts of Makefiles as much as possible.

The current implementation imposes a limitation: all software apps must have a unique name. It is not possible therefore to e.g. have a `dnn/gemm` and `blas/gemm`. This condition is currently satisfied. Note that the limitation can be removed, but we chose not to tackle this at the moment to simplify the implementation.


fischeti:
LGTM! The issue in https://github.com/pulp-platform/snitch_cluster/pull/147 was also resolved this way👍

fischeti:
Does the tutorial need to be updated?

colluca:
> Does the tutorial need to be updated?

I will follow with a tutorial update after merging https://github.com/pulp-platform/snitch_cluster/pull/71, since it also affects the benchmarking utilities.

PR #172, opened 2024-07-30
deps: Track `Bender.lock` file

fischeti:
A new AXI patch release broke the Gitlab-CI, and bender automatically incremented the AXI version since we don't have a `Bender.lock` file tracked that locks the IP versions used. To prevent similar problems in the future and to give users of the snitch cluster a clear set of dependency versions I suggest to track the `Bender.lock` file.

Apart from that, this PR does also:
* Bump the AXI version
* Remove the `Bender.local` file, which is redundant now and is anyway more of a hack.
* Fixes some minor issues for VCS simulation that were somehow triggered only after updating the dependencies.

### TODO:
- [ ] Use new AXI release (`0.39.5`) once it is published

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 
 # Dependency files
 /.bender/
-/Bender.lock
 
 # Modelsim Files
 *.wlf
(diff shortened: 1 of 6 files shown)


colluca (approved):

colluca:
Merging this to unblock development on other PRs. We can later bump the AXI version.

PR #173, opened 2024-08-02
Fix axi dependency in Bender

Lore0599:
The axi **vcs-fixes** branch has been merged into the **master** branch some days ago.
The [Bender file](https://github.com/pulp-platform/snitch_cluster/blob/4f55f01ffa4e99421af1b71b5bbbf867faf692bf/Bender.yml#L22) is still trying to point to that branch.
I change the revision to the last [axi commit](https://github.com/pulp-platform/axi/commit/4e54ac6766b160217a83a74d5a23af9bbf59e6ee).

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -19,7 +19,7 @@ package:
     - Matheus Cavalcante <matheusd@iis.ee.ethz.ch>
 
 dependencies:
-  axi:                { git: https://github.com/pulp-platform/axi,                rev:  vcs-fixes  }
+  axi:                { git: https://github.com/pulp-platform/axi,                rev:  4e54ac6766b160217a83a74d5a23af9bbf59e6ee  }
   axi_riscv_atomics:  { git: https://github.com/pulp-platform/axi_riscv_atomics,  version:  0.6.0   }
   common_cells:       { git: https://github.com/pulp-platform/common_cells,       version:  1.35.0  }
   FPnew:              { git: "https://github.com/pulp-platform/cvfpu.git",        rev:      pulp-v0.1.3 }


colluca:
I agree we should fix this but probably we can wait for a new AXI release, right? As long as `Bender.lock` points to an existing hash, this should not be an issue. Or are you experiencing errors?

Lore0599:
Yes I had issues with the chimera repo. Bender said that it couldn't satisfy the requirement 'vcs-fixes'.

colluca (approved):

fischeti:
> @fischeti do you think having removed the `Bender.local` file might be an issue? We thought it could simply be replaced by tracking the lock file, which is true for the Snitch cluster standalone, but I'm not sure when used as a dependency (for e.g. the reasons mentioned in this post), since I guess the lock file is not considered, whereas I believe the `Bender.local` file is.

I don't think tracking the `Bender.local` is the way to go. The [documentation](https://github.com/pulp-platform/bender?tab=readme-ov-file#package-structure) of Bender does not recommend it since it should only be used for _local_ development (since it is frankly kind of a hack). Regarding the `Bender.lock` file, it is true that this only works for locking the dependency for standalone development of the snitch cluster, but this is simply how it works. When you specify dependencies (incl. the snitch cluster), you need to resolve all potential version conflicts again to create a new `Bender.lock` file, since bender needs to have a unique version for each IP. The issue that @Lore0599 reported here was simply that the version specified in the `Bender.yml` did not exist anymore since it was merged/squashed in the AXI repo. But I don't think that the `Bender.local` file is part of the problem, since I am pretty sure that it will be ignored as well during version resolution.

PR #180, opened 2024-08-14
sw: Fix invalid inline assembly in `snrt_dma_wait_channel`

zero9178:
The inline assembly uses `t1` both as input operand (via the `cfg` variable) and within the clobber list. This is illegal but due to a bug in clang 12 that has been fixed in clang 14 has gone unnoticed. Quoting from the GCC documentation (https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers-1):

> Clobber descriptions may not in any way overlap with an input or output operand. For example, you may not have an operand describing a register class with one member when listing that register in the clobber list. Variables declared to live in specific registers (see Variables in Specified Registers) and used as asm input or output operands must have no part mentioned in the clobber description.

This correctly emits an error if using Clang 14 or any version of GCC: https://godbolt.org/z/17zrdK18e

The fix is to simply remove the clobber of `t1` which is safe to do as it is only used as input.


fischeti (approved):
Thanks! That went unnoticed since we are still using a very old Clang version😉

colluca (approved):
Thanks for the fix @zero9178 :)

PR #184, opened 2024-08-21
hw: Add debug information to error messages

colluca:
- Add misaligned address to the error message presented upon misaligned load/stores.
- Replace `$display` system function with `$info`, which already includes the timestamp as well as additional info, for misaligned load/stores and illegal instruction messages.


fischeti (approved):
LGTM, alternatively you could also use `$info` or even `$error` instead of `$display` which prints the simulation time as well (with a bit more verbosity).

colluca:
> LGTM, alternatively you could also use `$info` or even `$error` instead of `$display` which prints the simulation time as well (with a bit more verbosity).

That's a good point, this would probably be best changed everywhere. I guess we can keep this open until I have some time to address this.

PR #187, opened 2024-08-27
util: Replace `spike-dasm` in favor of `llvm-mc`

fischeti:
For trace generation we currently use `spike-dasm` to dissassemble single instructions from core log files. The problem with `spike-dasm` is that is an external executable, that needs to be maintained by us. For instance, new instructions need to be added manually to `spike-dasm` for decoding.

The better alternative is [`llvm-mc`](https://llvm.org/docs/CommandGuide/llvm-mc.html) which is a CLI tool that is compiled alongside LLVM. This means all our custom ISA extensions supported by the LLVM toolchain are also automatically supported by `llvm-mc`  and don't have to be added manually to the dissasembler as before.

### DIFF
--- a/iis-setup.sh
+++ b/iis-setup.sh
@@ -24,11 +24,3 @@ rm -rf tmp
 
 # Install local packages in editable mode.
 pip install -e .
-
-# Install spike-dasm
-mkdir tools/
-cd tools/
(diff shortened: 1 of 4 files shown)


colluca (changes requested):
LGTM apart from one minor comment :)

colluca (approved):

Issue #176, opened 2024-08-09
Update AXI dependency to new release when available


colluca:
Update both `Bender.lock` and `Bender.yml` files.

colluca:
Addressed in https://github.com/pulp-platform/snitch_cluster/pull/193.

PR #193, opened 2024-09-03
treewide: Miscellaneous changes to bump dependency in Occamy

colluca:
Implements several changes to bump the Snitch cluster dependency in Occamy:
- Renames the `snitch_cluster` Bender target, since this is now used by the DMA and required in Snitch-based systems, to `snitch_cluster_wrapper` containing the sources which are only needed to simulate/synthesize the `snitch_cluster` out of context
- Bump non-free repo which was updated with the change above
- Improve Makefile reusability in derived Snitch-based systems
- Remove all DWARF targets since these are unused
- Encode some missing linker prerequisites
- Several changes to improve reuse of Makefiles in derived systems (rename `ROOT` to `SNITCH_ROOT`, rename `PLATFORM_HEADERS` to `SNRT_HAL_HDRS`, rename `snitch_cluster_start.S` to `snrt.S`, make `MEMORY_LD` overridable)
- Several improvements to the Snitch runtime including accelerating all-to-all reduction function by using a hierarchical scheme, cores within a cluster reduce in TCDM.
- Generate accurate RTL dependencies using Verilator. The previous approach, already reported as fragile in https://github.com/pulp-platform/snitch_cluster/pull/175, was failing in Occamy, as blindly appending `/*/*` to all include directories would lead to "No rule to make target" errors.
- Move VCS targets to dedicated Makefile
- Remove `verible`, which was not correctly installed in container, and is anyways unused in Snitch
- Move `deploy-docs` job to CI workflow, to properly encode the  `build-docker`->`deploy-docs` dependency
- Address https://github.com/pulp-platform/snitch_cluster/issues/176


fischeti:
LGTM! What's the problem exactly with the overlapping `snitch_cluster` target in this repo and the DMA?

colluca:
> LGTM! What's the problem exactly with the overlapping `snitch_cluster` target in this repo and the DMA?

Basically the iDMA uses `snitch_cluster` as a Bender target for the Snitch DMA sources. We were using the same target for the `snitch_cluster_wrapper` sources. Occamy, and derived systems, need Snitch's iDMA sources and will thus need to include the `snitch_cluster` Bender target, but they should not include the wrapper (this failed to compile in Occamy). Thus we need distinct Bender targets.

PR #199, opened 2024-12-04
Feat: integer sorting example

nicoca20:
Added an integer Sorting algorithm, running on multiple cores in parallel.
Also added the signed type int32 in the data_utils file to the dictionary NP_DTYPE_FROM_CTYPE to test with negative integer numbers.


colluca:
Thanks @nicoca20, I can take it on from here.

colluca:
Superseded by https://github.com/pulp-platform/snitch_cluster/pull/294.

PR #203, opened 2025-02-14
Fix XBAR DMA address rules, add assertions

Scheremo:
This PR fixes a bug where the XBAR DMA's rules are picked to be the alias ruleset when `AliasRegionEnable = 1'b0`.
It further adds assertions of the `enabled_dma_xbar_rules` to ascertain that the first rules are always w.r.t absolute addresses.


colluca:
Thanks for the contribution @Scheremo!

I'm moving this contribution to https://github.com/pulp-platform/snitch_cluster/pull/204.

PR #218, opened 2025-04-10
Add official riscv-tests ISA tests as software tests

etterli:
This add a compile and test flow for the [riscv-tests](https://github.com/riscv-software-src/riscv-tests) ISA tests to the Snitch repo.

**Current limitation**
The float and double tests cannot be built with the Snitch LLVM toolchain as there is a special symbol to mark a float value. The error is how the float and double fields are initialized. A static float value is defined like:
```
fdata_label:
.float16 0f:7fc00000;
```
and the `0f:` prefix is unkown to the current Snitch assembler. The same applies to double values where the prefix is `0d:`.

For the actual code see `sw/deps/riscv-tests/isa/rv64uf/fadd.S:33` (i.e., macro `TEST_FP_OP_S_INTERNAL` at `sw/deps/riscv-tests/isa/macros/scalar/test_macros.h:449`) and the define of the `qNaNf` value at `sw/deps/riscv-tests/isa/macros/scalar/test_macros.h:422`.


etterli:
Limitations have been resolved due to updates to the riscv-tests repo.

This PR now integrates the tests into the test flow. However, certain tests for the F and D extension fail. Reason is, that the Snitch core does not order FPU instructions with CSR reads. Thus in certain test cases the `fcsr` is read before the relevant FPU instruction has finished, leading to a failing test case. The relevant test cases have been disabled for the time beeing.

target/snitch_cluster/sw/riscv-tests/riscv-tests.mk
```diff
@@ -0,0 +1,129 @@
...
+#
+# This Makefile builds the ISA tests from the RISC-V Test repository.
+# It makes use of the ETHZ LLVM clang infrastructure instead of the provided
+# gcc build setup in the original repository.
+# See also: https://github.com/riscv-software-src/riscv-tests
+#
+# It ignores following tests as we cannot build them with the snitch
+# compiler toolchain.
+# - rv32ua: specified immediate (rv32ua/../rv64ua/amomaxu_w.S:34:55) exceeds integer range
+# - rv32uf: The qNaN #define uses the '0f:' specifier which is not recognized by our LLVM.
```
colluca, line 15: I suppose this is no longer the case, is that correct? Double check also following two lines.

target/snitch_cluster/sw/riscv-tests/riscv-tests.mk
```diff
@@ -0,0 +1,129 @@
...
+#
+# It ignores the following test cases as we don't implement the extensions
+# - rv32uc
+# - rv32si
+# - rv32uzfh
+# - rv32uzba
+# - rv32uzbb
+# - rv32uzbc
+# - rv32uzbs
+
```
colluca, line 27: Move this entire comment block to the point where the Makefrags are included below, to make sure that the comment is always updated together with the include directives.

colluca (changes requested):

colluca (approved):

PR #225, opened 2025-04-22
Replace reggen and hjson with rdl

micprog:


hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
```diff
@@ -0,0 +1,242 @@
...
+            desc = "Incremented for instruction cache double hit.
+                    _This is a hart-local signal_";
+        };
+        ICACHE_STALL = 30 {
+            desc = "Incremented for instruction cache stalls.
+                    _This is a hart-local signal_";
+        };
+    };
+
+    reg PERF_CNT_SEL {
```
fischeti, line 162: Any chance we can define reset values for array regs individually? Then we don't even have to define `PERF_CNT_SEL` as external registers. The idea is that the performance counters will start tracking a couple of different metrics at start up.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
```diff
@@ -0,0 +1,242 @@
...
+        desc = "Controls prefetching of the instruction cache.";
+        field {
+            desc = "Enable instruction prefetching.";
+            hw = r;
+            sw = w;
+        } ICACHE_PREFETCH_ENABLE[0:0];
+    };
+
+             PERF_CNT_EN    PERF_CNT_EN[NumPerfCounters]   @0x0   += 0x8;
+    external PERF_CNT_SEL   PERF_CNT_SEL[NumPerfCounters]  @0x100 += 0x8;
```
fischeti, line 234: Is it possible to just align the address to the previous regfile? Not a big problem, just curious

hw/snitch_cluster/src/snitch_cluster_wrapper.rdl.tpl
```diff
@@ -0,0 +1,44 @@
...
+        memwidth = ${cfg['cluster']['data_width']};
+    };
+
+    mem ZEROMEM {
+        mementries = ${hex(int(cfg['cluster']['zero_mem_size'] * 1024 * 8 / cfg['cluster']['data_width']))};
+        memwidth = ${cfg['cluster']['data_width']};
+    };
+
+
+    external TCDM                 TCDM           @BASE_ADDR;
```
fischeti, line 34: what does `external` mean in this context?

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv
```diff
@@ -43,20 +47,41 @@ module snitch_cluster_peripheral
...
-    .reg2hw (reg2hw),
-    .hw2reg (hw2reg)
+  snitch_cluster_peripheral_reg__out_t reg2hw;
+  snitch_cluster_peripheral_reg__in_t  hw2reg;
+
+  `APB_TYPEDEF_ALL(sn_periph_regs_apb, addr_t, data_t, strb_t)
+  sn_periph_regs_apb_req_t  sn_periph_regs_apb_req;
+  sn_periph_regs_apb_resp_t sn_periph_regs_apb_rsp;
+
+  reg_to_apb #(
```
fischeti, line 57: I guess we could go directly to APB from AXI in the cluster level

hw/snitch_cluster/src/snitch_cluster_wrapper.rdl.tpl
micprog: `external` refers to the register being described external to the register block, i.e., requiring manual definition: https://peakrdl-regblock.readthedocs.io/en/latest/rdl_features/external.html#external-registers
From my understanding, it is similar to the `hwext` specification in the opentitan hjson format.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
micprog: I don't think there is a straightforward way to specify this. Currently, these registers are specified as external, so the reset values should be applied externally, i.e., here: https://github.com/pulp-platform/snitch_cluster/blob/main/hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv#L202-L209

fischeti: The reason why they were defined as externally was because reset values cannot be defined individually with the regtool🤓

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
fischeti: I see👍 I don't think the actual address is interesting in the end as long as we reference the generated files. I think the `compact` statement just means that multiple fields can be bundled into a 32-bit register (to save some address space I guess). But I don't think we actually require it.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv
fischeti: I would use APB😉

colluca:
Simple question, why? :grin:

colluca:
> SystemRDL together with PeakRDL provides great infrastructure for register files, and especially can tie together these address maps hierarchically, allowing for a full view of an SoC. The language is properly specified and more widely accepted than opentitan's hjson format.

I see now that it's an Accelera standard. Cool, thanks @micprog :)

fischeti:
Currently, some tests are failing because the BSS section in the tests is very large. The `.bss` section is initialized in the beginning with the DMA and the zero memory. But if the BSS section is too large, the DMA transfers will overflow the zero memory address range. It can be fixed in SW, but since we anyway will get rid of the zero memory in https://github.com/pulp-platform/snitch_cluster/pull/238 it might make more sense to wait for this PR to be merged.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
```diff
@@ -0,0 +1,255 @@
+// Copyright 2025 ETH Zurich and University of Bologna.
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+// Licensed under Solderpad Hardware License, Version 0.51, see LICENSE for details.
```
colluca, line 4: Redundant line

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
```diff
@@ -0,0 +1,255 @@
...
+    };
+
+    reg unused {
+        desc = "Unused register added to print enum";
+        field {
+            sw=r; hw=na;
+            encode = PERF_METRIC;
+        } f[4:0] = 0;
+    };
+
```
colluca, line 240: What is this for exactly?

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv
```diff
@@ -43,100 +46,84 @@ module snitch_cluster_peripheral
...
   always_comb begin
     cl_clint_d = cl_clint_q;
-    if (reg2hw.cl_clint_set.qe) begin
-      cl_clint_d = cl_clint_q | reg2hw.cl_clint_set.q;
-    end else if (reg2hw.cl_clint_clear.qe) begin
-      cl_clint_d = cl_clint_q & ~reg2hw.cl_clint_clear.q;
+    if (reg2hw.CL_CLINT_SET.req && reg2hw.CL_CLINT_SET.req_is_wr) begin
+      cl_clint_d = cl_clint_q | reg2hw.CL_CLINT_SET.wr_biten.CL_CLINT_SET &
+                                reg2hw.CL_CLINT_SET.wr_data.CL_CLINT_SET;
+    end else if (reg2hw.CL_CLINT_CLEAR.req && reg2hw.CL_CLINT_CLEAR.req_is_wr) begin
```
colluca, line 94: Parenthesis around the & operation, for clarity.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
colluca, line 1: Would lowercase names here resolve all the mixed case names in the generated RTL and C sources? Would seem to be also more in line with the style guide https://peakrdl.readthedocs.io/en/latest/style-guide.html#component-type-and-instance-names-are-lowercase

target/common/rtl.mk
```diff
@@ -8,26 +8,33 @@ SN_BOOTROM_DIR ?= $(SN_TARGET_DIR)/test
 # Templates
 SN_CLUSTER_WRAPPER_TPL = $(SN_HW_DIR)/snitch_cluster/src/snitch_cluster_wrapper.sv.tpl
 SN_CLUSTER_PKG_TPL     = $(SN_HW_DIR)/snitch_cluster/src/snitch_cluster_pkg.sv.tpl
+SN_CLUSTER_RDL_TPL	   = $(SN_HW_DIR)/snitch_cluster/src/snitch_cluster_wrapper.rdl.tpl
 
 # Generated RTL sources
 SN_CLUSTER_WRAPPER    = $(SN_GEN_DIR)/snitch_cluster_wrapper.sv
 SN_CLUSTER_PKG        = $(SN_GEN_DIR)/snitch_cluster_pkg.sv
-SN_CLUSTER_PERIPH_TOP = $(SN_PERIPH_DIR)/snitch_cluster_peripheral_reg_top.sv
+SN_CLUSTER_PERIPH_TOP = $(SN_PERIPH_DIR)/snitch_cluster_peripheral_reg.sv
```
colluca, line 16: Drop `TOP` as misleading

target/common/rtl.mk
```diff
@@ -8,26 +8,33 @@ SN_BOOTROM_DIR ?= $(SN_TARGET_DIR)/test
...
 SN_CLUSTER_PERIPH_PKG = $(SN_PERIPH_DIR)/snitch_cluster_peripheral_reg_pkg.sv
 SN_BOOTROM            = $(SN_BOOTROM_DIR)/snitch_bootrom.sv
+SN_CLUSTER_RDL        = $(SN_GEN_DIR)/snitch_cluster_wrapper.rdl
+SN_CLUSTER_ADDR_SVH   = $(SN_GEN_DIR)/snitch_cluster_addrmap.svh
```
colluca, line 20: Is this used anywhere? Otherwise remove

target/common/rtl.mk
```diff
@@ -8,26 +8,33 @@ SN_BOOTROM_DIR ?= $(SN_TARGET_DIR)/test
...
 $(SN_CLUSTER_PERIPH_PKG): $(SN_CLUSTER_PERIPH_TOP)
-$(SN_CLUSTER_PERIPH_TOP): $(SN_PERIPH_DIR)/snitch_cluster_peripheral_reg.hjson
-	@echo "[REGGEN] Generating peripheral regfile"
-	$(REGGEN) -r -t $(SN_PERIPH_DIR) $<
+$(SN_CLUSTER_PERIPH_TOP): $(SN_PERIPH_DIR)/snitch_cluster_peripheral_reg.rdl
+	@echo "[peakrdl] Generating peripheral regfile"
+	$(PEAKRDL) regblock $< -o $(PERIPH_DIR) --cpuif apb4-flat --default-reset arst_n
+$(SN_CLUSTER_ADDR_PKG): $(SN_CLUSTER_RDL)
+	@echo "[peakrdl] Generating addr pkg"
+	$(PEAKRDL) raw-header $< -o $(SN_GEN_DIR) --format svh -I $(SN_PERIPH_DIR)
```
colluca, line 38: Is this used anywhere? `SN_CLUSTER_ADDR_PKG` does not seem to be defined here

colluca (changes requested):
Thanks for the contribution @micprog! It mostly LGTM, I just have some minor questions for you in the following review. Regarding the suggested changes I could also apply these myself.

.github/workflows/lint.yml
```diff
@@ -67,6 +67,9 @@ jobs:
           match_regex: true
           exclude_paths: |
             sw/snRuntime/src/omp/interface.h
+            hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg_pkg.sv
+            hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.sv
+            hw/snitch_cluster/src/snitch_cluster_wrapper.rdl.tpl
```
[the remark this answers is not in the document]
micprog: The license is there in the file, but as the license script doesn't properly understand the rdl format (and assumes this file ending instead of tpl because of the tpl ending), it doesn't get handled correctly.

pyproject.toml
```diff
@@ -33,6 +33,9 @@ dependencies = [
     "mkdocstrings-python",
     "numpy",
     "pandas",
+    "peakrdl",
+    "peakrdl-markdown",
+    "peakrdl-rawheader@git+https://github.com/micprog/PeakRDL-rawheader.git",
```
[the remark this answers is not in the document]
micprog: The `raw-header` plugin allows dumping of a more raw format (i.e., just addresses and offsets) as either a C header, systemverilog header, or systemverilog package. The way peakrdl works is that anyone can make a custom plugin, so this is the upstream version (it is just not published to pip). We can lock it to a hash if you want, or I can also add version tags. This is needed for a systemverilog header, which is needed by the testbench that needs the addresses. The regblock package does not (and will not https://github.com/SystemRDL/PeakRDL-regblock/pull/139) expose these addresses or offsets as usable parameters.

Bender.lock
```diff
@@ -94,7 +94,7 @@ packages:
     revision: 5daa85d164cf6b54ad061ea1e4c6f3624556e467
     version: 0.4.5
     source:
-      Git: https://github.com/pulp-platform/register_interface
+      Git: https://github.com/pulp-platform/register_interface.git
```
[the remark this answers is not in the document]
micprog: I think it's a subdependency of idma. AFAIK @fischeti removed it from the Bender.yml.

target/snitch_cluster/sw/runtime/common/snitch_cluster_memory.h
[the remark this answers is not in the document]
micprog: Agreed, great item for a future PR 🤷

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
colluca: I converted all names to lowercase as mentioned.

hw/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral_reg.rdl
colluca: Thanks for clarifying, I added a comment to reflect this.

PR #226, opened 2025-04-29
Eliminate PIP cache and force torch cpu only install

uge:
The current docker container is ~12.5GB - a lot of which is nvidia stuff brought in by torch and pip cache. This change modifies the dockerfile to explicitly install the cpu only version of torch and disables the PIP cache to shrink the docker image from 12.5GB to about 4.5GB.


colluca (approved):
Thanks for the contribution @uge!

PR #227, opened 2025-05-06
treewide: Improve integration in systems

colluca:
Several changes to streamline integration in systems, tested in [Picobello](https://github.com/pulp-platform/picobello/pull/29).

Key features:
- Better separation of Make fragments: `vcs.mk`, `vsim.mk`, `verilator.mk`, `rtl.mk`
- Prefix Makefile variables with `SN_` and targets with `sn-` to prevent name collisions when included in systems

Note: this PR is only a first in this direction, a more complete and consistent implementation will be merged shortly with a second PR.

### DIFF
--- a/target/snitch_cluster/sw/tests/tests.mk
+++ b/target/snitch_cluster/sw/tests/tests.mk
@@ -8,69 +8,66 @@
 # Directories #
 ###############
 
-TESTS_SRCDIR   = $(ROOT)/sw/tests
-TESTS_BUILDDIR = $(ROOT)/target/snitch_cluster/sw/tests/build
+SNRT_TESTS_SRCDIR    = $(SN_ROOT)/sw/tests
+SNRT_TESTS_BUILDDIR ?= $(SN_ROOT)/target/snitch_cluster/sw/tests/build
 
 ###################
 # Build variables #
 ###################
 
-TESTS_RISCV_CFLAGS += $(RISCV_CFLAGS)
-TESTS_RISCV_CFLAGS += $(addprefix -I,$(SNRT_INCDIRS))
+SNRT_TESTS_RISCV_CFLAGS += $(RISCV_CFLAGS)
+SNRT_TESTS_RISCV_CFLAGS += $(addprefix -I,$(SNRT_INCDIRS))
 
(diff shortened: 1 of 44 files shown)


colluca:
> @colluca I think that some of the targets in the CI need to be modified. For instance [here](https://github.com/pulp-platform/snitch_cluster/blob/05ed6e4c41a5dc54d7b2b2d34f87eb188cb7f52f/.gitlab-ci.yml#L69) the CI should probably run `sn-sw`.

Definitely yes, it's really still just a draft, I want to prefix the other targets and variables as well.

fischeti (approved):
LGTM! Thanks a lot

target/snitch_cluster/sw/tests/tests.mk
fischeti, line 1: I just had a question about the tests. If someone wants to add other snitch tests in external repositories. How could they be added the best way? E.g. see https://github.com/pulp-platform/picobello/pull/27 from @mbertuletti. Would it be possible to just add additional tests to `SNRT_TEST_NAMES` when including `sw.mk`?

colluca: Unfortunately not, although that would be very nice.

The reason is the rules are defined as:
```Makefile
$(SNRT_TESTS_BUILDDIR)/%.elf: $(SNRT_TESTS_SRCDIR)/%.c $(SNRT_LD_DEPS) $(SNRT_TESTS_BUILDDIR)/%.d | $(SNRT_TESTS_BUILDDIR)
	$(RISCV_CC) $(SNRT_TESTS_RISCV_CFLAGS) $(SNRT_TESTS_RISCV_LDFLAGS) $(SNRT_TESTS_SRCDIR)/$*.c -o $@
```

So any file not contained within `$(SNRT_TESTS_SRCDIR)` won't match.
The problem fundamentally is that you want sources from different directories to compile to binaries in the same directory, and we are in no way telling it about multiple source directories.

This can be improved, and I will anyways streamline the whole software build flow when I manage to get there, but for me the best approach now (and maybe in general) would be to just point `$(SNRT_TESTS_SRCDIR)` to some directory in Picobello, put all external Snitch tests there, and if you want any specific test from Snitch, symlink it there.

PR #231, opened 2025-05-16
Prevent unaligned memory accesses

and-ivanov:
Fixes unaligned memory access errors:
```
float *SLASHfc1SLASHGemm_output_0;
       SLASHfc1SLASHGemm_output_0 = (float *) snrt_l3_alloc(sizeof(float ) * (4));
...
SLASHfc1SLASHGemm_output_0[((2 * i) + j)] = o;
```
Error:
```
[1358000] -Info: snitch.sv:2934: TOP.testharness.i_snitch_cluster.i_cluster.gen_core[0].i_snitch_cc.i_snitch: [Misaligned Load/Store Core 0] PC: 000080000280 Data: 00252027 Addr: 8000c679
```

### DIFF
--- a/sw/snRuntime/src/alloc.h
+++ b/sw/snRuntime/src/alloc.h
@@ -67,6 +67,8 @@ inline void snrt_l1_update_next(void *next) {
 inline void *snrt_l3_alloc(size_t size) {
     snrt_allocator_t *alloc = snrt_l3_allocator();
 
+    size = ALIGN_UP(size, MIN_CHUNK_SIZE);
+    
     // TODO: L3 alloc size check
 
     void *ret = (void *)alloc->next;


colluca:
Hi @and-ivanov, thanks for the contribution :)

Could you enable the CI on your fork, so we can check that the other jobs also pass?

Also, maybe you want to look into the `alloc_v2.h` functions, I find them more convenient to work with then the `alloc.h` functions. But we could also have a quick talk about this perhaps.

colluca:
> Hi @and-ivanov, thanks for the contribution :)
> 
> Could you enable the CI on your fork, so we can check that the other jobs also pass?
> 
> Also, maybe you want to look into the `alloc_v2.h` functions, I find them more convenient to work with then the `alloc.h` functions. But we could also have a quick talk about this perhaps.

About the v2 functions, I just saw in your other PR that you already know about them :grin:

colluca:
> `alloc_v2.h` seem to target only L1 allocations, this is the fix for L3. I realized you require clang-format over sources. I also enabled actions in the fork but they will probably appear only in the future commits.

Yes, indeed. Regarding the CI you can just trigger it by `git commit --amend` and `git push --force`, even if you don't make any actual changes to the commit.

PR #232, opened 2025-05-16
Make Snitch runtime compatible with C++

and-ivanov:
Fix compilation errors on inclusion of Snitch runtime headers into .cpp code by making type casting explicit and avoiding arithmetics on void* pointers.

### DIFF
--- a/sw/snRuntime/src/alloc_v2.h
+++ b/sw/snRuntime/src/alloc_v2.h
@@ -83,7 +83,7 @@ inline void *snrt_l1_alloc_compute_core_local(size_t size,
                                               const size_t alignment) {
     snrt_l1_allocator_v2()->next =
         ALIGN_UP(snrt_l1_allocator_v2()->next, alignment);
-    void *retval = snrt_l1_next_v2() + size * snrt_cluster_core_idx();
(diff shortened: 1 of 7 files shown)


sw/snRuntime/src/alloc_v2.h
```diff
@@ -83,7 +83,7 @@ inline void *snrt_l1_alloc_compute_core_local(size_t size,
                                               const size_t alignment) {
     snrt_l1_allocator_v2()->next =
         ALIGN_UP(snrt_l1_allocator_v2()->next, alignment);
-    void *retval = snrt_l1_next_v2() + size * snrt_cluster_core_idx();
+    void *retval = ((uint8_t*) snrt_l1_next_v2()) + size * snrt_cluster_core_idx();
```
colluca, line 86: Perhaps it could be even better to just return `uintptr_t` types in the alloc functions, so the arithmetic can be performed on that type, and then cast (only once) to whatever pointer one may need. What do you think? Would that work for your use case?

and-ivanov: I don't have strong opinion but it will go against the "malloc" convention

```
     void *malloc(size_t size);
```

colluca (approved):

PR #235, opened 2025-05-28
hw(periph): Convert from AXI directly to APB

fischeti:


micprog (approved):
LGTM, if register_interface is no longer used in the cluster it can probably also be removed from he Bender.yml (and Bender.lock)

fischeti:
> LGTM, if register_interface is no longer used in the cluster it can probably also be removed from he Bender.yml (and Bender.lock)

jep, already had it mind🤓 it stays in `Bender.lock` though since it is a sub-dependency

Issue #88, opened 2024-02-05
make bin/snitch_cluster.vlt error


ashuthosh-mr:
I have built the docker and cloned the snitch_cluster repo in /repo. 
And then in the target/snitch_cluster, I use the command make bin/snitch_cluster.vlt

For which I get: 

 work-vlt/Vtestharness.h:11:10: fatal error: verilated_heavy.h: No such file or directory

What am I doing wrong here?

Commands I have used:
1. In the snitch_cluster root, docker run -it -v $REPO_TOP:/repo -w /repo ghcr.io/pulp-platform/snitch_cluster:main
2. In /repo of docker, git clone https://github.com/pulp-platform/snitch_cluster.git --recurse-submodules
3. In /repo/snitch_cluster/target/snitch_cluster,  make bin/snitch_cluster.vlt

Thanks in advance.

ashuthosh-mr:
I changed the Makefile, common.mk and run.py to previous version of the update #78 in the /target/snitch_cluster/ and it worked this time. I am able to run an application i.e axpy.elf. I will update more on this issue.

ashuthosh-mr:
I faced further issues while running an application that had floating division (illegal instruction error). After going through the dnn examples that were part of the repository, I found that to make it work I need to enable divsqrt. Everything works fine now, after enabling divsqrt in the config json file. 

Thanks, I shall close this issue.

PR #238, opened 2025-05-30
idma: Streamline TCDM connection, enable multi-channel operation

thommythomaso:
Modify the iDMA integration into the Snitch cluster.

This includes:
* The wide port from the SoC directly connects to the TCDM subsystem, allowing external data access while DMA transfers run
* The wide DMA crossbar has been simplified 
* The DMA uses the OBI protocol to connect to the TCDM subsystem
* Arbitration between the channels, the superbanks, and the SoC port happens now in TCDM, increasing flexibility and throughput. With minimal changes, it is possible to have multiple DMA cores in the cluster. 

Before merging:
* [ ] Update `custom_instructions.md`
* [ ] Fix failing tests. 
* [ ] Fix linting issues.
* [ ] Check for (*perfomrance*) regressions
* [ ] Merge the corresponding [branch](https://github.com/pulp-platform/iDMA/tree/luzil/interl) in the *iDMA* repo into main [#88](https://github.com/pulp-platform/iDMA/pull/88)
* [x] Merge the corresponding [branch](https://github.com/LuziLucia/riscv-opcodes/tree/master) in the *riscv-opcodes* repo into main


fischeti:
Some notes here on current backend trials I am doing with this:

* The `wide_in` path should be cut (by default). At the moment this results in large interface timing violations since there is a combinational path from TCDM to the AXI interface now (and vice-versa). There is already a parameter `RegisterExtWide` to cut those paths. However, the same parameter is also used to cut the `wide_out`, which is not necessary since that path already has a cut from the wide AXI Xbar. It would therefore make sense to differentiate with `RegisterExtWideIn` and `RegisterExtWideOut`.
* The TCDM interconnect grows quite a bit with a configuration of 4 DMA channels. This could also be due to the additional external TCDM port, that currently cannot be parametrized away.

I would continue with trials, but I think it makes more sense to wait for https://github.com/pulp-platform/snitch_cluster/pull/289 to be merged and then rebase this PR.

colluca:
Superseded by https://github.com/pulp-platform/snitch_cluster/pull/322.

PR #241, opened 2025-06-04
Updated iDMA tests

RiccardoGandolfi:
Added the following tests for iDMA instance in snitch cluster:

- idma_1d: executes 1D transfers on iDMA (rework of idma_simple to work at byte granularity)
- idma_2d: executes 2D transfers on iDMA


sw/tests/dma_2d.c
colluca, line 1: Same comments as on `dma_1d.c`. Double check also pointers in printf statements.

colluca:
@RiccardoGandolfi On a more general note, do you think there's still a benefit in keeping the `dma_simple` test? It seems to me that your `dma_1d` is a generalization of it, so perhaps we can get rid of the old one, to reduce duplication.

PR #242, opened 2025-06-06
Add contribution guidelines

colluca:


Bender.yml
fischeti, line 1: Just a minor comment. I would add `*.git` to every url. Then bender will not interpret it as different remotes and dependency resolving is easier

fischeti (commented):
Good idea to setup guidelines 👍 Is it on purpose that all the multicast stuff is also part of this PR?

PR #257, opened 2025-07-08
use https in gitmodules

superlopuh:


colluca:
Hi @superlopuh,

To  my knowledge the SSH-based sign in is actually desired. We switched to it when Github removed password-based authentication, which is the default sign in mechanism with the HTTPS URLs. Is there a reason to prefer HTTPS over SSH?

superlopuh:
CC @compor This was one of the things we had to change for our docker builds to work. I agree that ssh is preferable for authentication when pushing to the repo, but it's more limiting when using it read-only, such as when building the docker. Maybe we'll just keep the current flow of string-replacing ssh for https when building the docker container.

colluca:
I see the problem, perhaps something like this could also work?

https://gist.github.com/Kovrinic/ea5e7123ab5c97d451804ea222ecd78a

superlopuh:
Oh cool, so the idea would be to add this kind of config as a build step in the docker instead of string matching?

PR #258, opened 2025-07-10
Expose multiple wide TCDM ports

belanoa:
This PR enables support for the exposure of multiple wide TCDM ports, useful if one wants to use multiple external accelerators at the same time.


colluca (changes requested):
Thanks for the contribution @belanoa! Please find my review below.

PR #266, opened 2025-07-26
Fix roi target

etterli:
The ROI make target used the old `CFG` variable to pass the current hardware configuration.
It seems that this variable has been renamed to `SN_CFG`


etterli:
It seems that the develop branch is one commit behind the main branch at the time of creating this PR..

colluca:
> It seems that the develop branch is one commit behind the main branch at the time of creating this PR..

Good point. Bumped it now.

colluca (approved):

PR #267, opened 2025-07-26
Replace non-synthesizable code in `snitch_cluster.sv`

belanoa:
In `cluster_tile.sv` the `dma_xbar_default_port ` variable is currently assigned on declaration, causing incorrect behaviors during synthesis.
This PR fixes the issue.

### DIFF
--- a/hw/snitch_cluster/src/snitch_cluster.sv
+++ b/hw/snitch_cluster/src/snitch_cluster.sv
@@ -659,7 +659,9 @@ module snitch_cluster
   );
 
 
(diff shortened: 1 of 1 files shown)


fischeti (commented):
I agree that this is the better way to write it in general. But I am suprised that this causes problems during synthesis, since this is very old code, and as far as I know did not cause problems previously. What tool are you using for synthesis?

Also, there are a couple others of those kind of statements in `snitch_cluster.sv`. Maybe you could change them as well.

Lastly, is this problem related to picobello? I.e. do we need to bump the snitch cluster?

PR #317, opened 2026-05-01
hw: Always offload unsupported instructions to CV-X-IF

colluca:
When the CV-X-IF is present, unsupported instructions don't raise illegal instruction, but are offloaded on the CV-X-IF.

Illegal instruction is only raised if no CV-X-IF coprocessor can handle it.


colluca:
Included in https://github.com/pulp-platform/snitch_cluster/pull/318.

PR #324, opened 2026-07-07
sw(deps): include submodule fragments conditionally

fischeti:
Adds a new `sn-sw-submodules` target in the software build flow to update the submodules on demand.

When snitch cluster is used a s a bender dependency in another system, bender until now automatically cloned all the submmodules as well. However, this can be very inefficient since submodules might not actually be used in downstream project (e.g. the `ihp130-pdk` submodule). Hence, bender v0.32.1 has gained support to disable submodule clones https://github.com/pulp-platform/bender/pull/314, and the plan is that future versions will by default not clone submodules at all unless specified in `Bender.yml` (see https://github.com/pulp-platform/bender/pull/318).

This new target allows to pull in the submodules on demand only.

Note: this PR is deliberately not rebased on to `develop`, since it apparently currently brakes in gwaihir


fischeti:
Closing this, since this can also be achieved on the integration side, see https://github.com/pulp-platform/gwaihir/pull/14

PR #326, opened 2026-07-22
Integrate Spatz in Snitch cluster

DiyouS:
**HW**

- Clean up core complex
  - Separate CC sources from snitch_cluster subdir to snitch_cc subdir
  - Move tracer logic to `snitch_tracer` module
  - Move SSR subsystem logic to `snitch_ssr_subsystem` module
  - Move address-based reqrsp demux logic to new `reqrsp_demux_mapped` module
- Clean up interfaces
  - Legacy `reqrsp_interface` interface was effectively an LSU interface. Separate and rename this interface to `lsu_interface`.
  - `reqrsp_interface` proper is now an abstract interface, from which different  concrete reqrsp interfaces can be created (e.g. `dca_interface`, `fpu_interface`, `lsu_interface`, etc.)
  - Generalize reqrsp IPs, VIPs and TBs to accept/test a generic reqrsp interface
  - Implement `tcdm_interface` and `mem_interface` based entirely on generic `reqrsp_interface`
- Conditionally add Spatz into core complex
  - XIF from core is demuxed to Spatz and external coprocessor port
  - TCDM ports are appended to CC's already existing TCDM ports
  - Spatz LSU port is muxed with Snitch and FPSS LSUs
- Rename Xf16 to Zfh extension everywhere
- Move `hart_id_t` from `snitch_cluster_pkg` to `snitch_pkg`
- Rename `riscv_instr` to `snitch_riscv_instr` to prevent name conflicts
- Separate `reqrsp_interface` and LSU IP/interface to dedicated repos. Allows to break the circular snitch_cluster → spatz → snitch_cluster(for LSU and reqrsp_interface) dependency, which now becomes snitch_cluster → spatz → LSU/reqrsp_interface

**SW**

- Add minimal Spatz test
- Add Spatz kernels
- Rename Xf16 to Zfh extension everywhere
- Only perform inter-cluster reduction in `snrt_global_all_to_all_reduction` if there is more than one cluster
- Add support for uint16 types to `data_utils`
- Add missing `SNRT_SUPPORTS_SMALLFLOAT` guard on naive FP16 GEMM

**Make**

- Replace `DEBUG=ON` for enabling traces with `TRACE=OFF` for disabling them (on by default). This allows to run performance experiments without having to run the simulation with `+acc`.
- Template `toolchain.mk` and derive toolchain flags directly from cluster configuration (`SN_MCPU` flag thus becomes obsolete)
  - This requires to break the Makefile read phase into multiple explicit stages, e.g. `toolchain.mk` has to have been built and read before the bootrom generation targets can be evaluated
- Update system integration docs to reflect `toolchain.mk` addition
- Redirect Verilator build stderr (in addition to stdout) output to log file
- Create separate Make goal for Verilator filelist
- Split bootrom goal into separate subgoals

**CI**

- Waive `explicit-parameter-storage-type` in `riscv_instr.sv`
- Upload Yosys artifacts on CI job failure
- Print calculated base SHA in CI for simplified debugging
- Fix `base-sha` job to pick up the upstream main branch even on forks
- Bump `make-prerequisites-changed` action to fix undetected file changes (see https://github.com/colluca/make-prerequisites-changed/pull/1)

**Misc**

- Checkout Bender dependencies before Python setup (Spatz needs to be checked out before its Python dependencies can be installed)
- Replace VCD with FST tracing in Verilator and enable tracing only if DEBUG=ON
- Add agentic waveform debugging flow based on tsunami
  - Bump required Python to >=3.12

**TODO**
- [x] Add Spatz configuration and tests to CI
- [x] Test in Gwaihir
- [x] Create releases for `reqrsp_interface` and `core_ips` repos under `pulp-platform` namespace
- [ ] ~Bump to use new toolchain [22.1.7-pulp-0.2.0](https://github.com/pulp-platform/llvm-project/releases/tag/22.1.7-pulp-0.2.0)~ (there are issues with the deployment CI, so do this in a separate PR)


cfg/spatz.json
```diff
@@ -0,0 +1,141 @@
...
+        n_ipu: 1,
+        n_fpu: 8,
+        vlen: 512,
+        spatz_nports: 8,
+        double_bw: false,
+        spatz_fpu: true,
+        mempool: false,
+        buf_fpu: 1,
+        rvf: true,
+        rvd: true,
```
colluca: Intuitively, these should be parameters of the core/CC, not of the cluster. Potentially, this would also allow us to have multiple CCs with different Spatz configurations.

hw/snitch/src/snitch.sv
```diff
@@ -3761,7 +3792,7 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
     .tag_t (logic[RegWidth-1:0]),
     .NumOutstandingMem (NumIntOutstandingMem),
     .NumOutstandingLoads (NumIntOutstandingLoads),
-    .Caq (FP_EN),
+    .Caq (1'b1),
```
colluca, line 3826: Why this change?

hw/snitch/src/snitch.sv
```diff
@@ -3468,6 +3480,25 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
...
+                 hart_id_i, $time,
+                 x_register_o.id,
+                 x_register_o.rs[2], x_register_o.rs[1], x_register_o.rs[0],
+                 x_register_o.rs_valid, x_register_ready_i);
+      if (x_result_valid_i)
+        $display("[XIF Core %0d] %t RESULT  id=%0d rd=%0d data=0x%08h we=%0b ready=%0b",
+                 hart_id_i, $time,
+                 x_result_i.id, x_result_i.rd, x_result_i.data,
+                 x_result_i.we, x_result_ready_o);
+    end
```
colluca: I would remove the custom tracing logic from this file, and instead add the instructions offloaded on the X-interface to the tracing infrastructure we already have in place.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -327,7 +327,14 @@ module snitch_cluster
   // [0:0]           is_core
   localparam int unsigned TcdmUserWidth = CoreIDWidth + 1;
 
+`ifdef DOUBLE_BW
```
colluca: Instead of a macro, this should preferably be a configuration parameter.

make/common.mk
```diff
@@ -54,7 +54,7 @@ SN_BENDER_LOCK = $(SN_ROOT)/Bender.lock
 SN_BENDER_YML  = $(SN_ROOT)/Bender.yml
 
 # Flags
-SN_COMMON_BENDER_FLAGS      += -t rtl -t cc_no_deprecated -t tech_cells_generic_include_tc_sync
+SN_COMMON_BENDER_FLAGS      += -t rtl -t cc_no_deprecated -t tech_cells_generic_include_tc_sync -t spatz
```
colluca: Why is this flag needed? Generally, you just include an IP as a dependency and don't need to add any target flag. I don't see the use case that requires this flag.

make/rtl.mk
```diff
@@ -30,6 +35,25 @@ SN_BOOTROM_ARTIFACTS = $(SN_BOOTROM_ELF) $(SN_BOOTROM_DUMP) $(SN_BOOTROM_BIN)
...
 
+# Spatz package generation via the Spatz hw/Makefile flow:
+#   import_cfg.py extracts Spatz VFU fields from the cluster config,
+#   apply_cfg.py renders spatz_pkg.sv.tpl into the generated file.
+$(SPATZ_PKG): $(SN_CFG) $(SPATZ_PKG_TPL) | $(SN_GEN_DIR)
+	@echo "[SPATZ] Generating $@"
+	cd $(SPATZ_HW_DIR) && python import_cfg.py $(abspath $(SN_CFG))
+	cd $(SPATZ_HW_DIR) && python apply_cfg.py $(basename $(notdir $(SN_CFG)))
+	cp $(SPATZ_HW_DIR)/src/generated/spatz_pkg.sv $@
+
```
colluca: Ideally, one would extend the generation script to:
1. enable invoking from any working directory, and
2. override the path where the output is to be generated.

These are merely suggestions, not strictly needed for this PR.
But particularly the latter is a safety feature that can avoid future headaches, e.g. if you're working on Spatz and the cluster simultaneously, and don't want that reconfiguring Spatz in the cluster repo overwrites the file in the Spatz repo, potentially affecting the parallel work on Spatz.

make/vsim.mk
```diff
@@ -26,6 +26,8 @@ SN_VLOG_FLAGS += -suppress 13314
 SN_VLOG_FLAGS += -work $(SN_VSIM_BUILDDIR)
 SN_VOPT_FLAGS += -work $(SN_VSIM_BUILDDIR)
 SN_VSIM_FLAGS += -64
+SN_VSIM_FLAGS += -suppress vsim-8386
+SN_VSIM_FLAGS += -suppress vsim-3999
```
colluca: I suppose these can't be fixed at the source?

sw/runtime/src/start.c
```diff
@@ -137,6 +137,12 @@ extern void snrt_exit(int exit_code);
 #endif
 #endif
 
+// Declared at file scope so it has C++ linkage (not C linkage) when compiled
+// as C++. Clang 22+ warns if this declaration appears inside extern "C" scope.
+#ifdef SNRT_INVOKE_MAIN
+extern int main();
```
colluca: Double check if necessary, I ended up not including this in the upstream commit which already switched to LLVM 22.

colluca (commented):
Apart from the minor comments/changes that follow, I think we should make one more major change, specifically in the integration of the software.

While this could potentially become a huge effort, to keep it reasonable I would aim only for the following two points:
1. Instead of a separate `sw/spatz` folder, move its contents respectively to the `sw/runtime` directory and to the subdirectories of `sw/kernels`. If a specific kernel already exists, there is no need to merge the original and the Spatz version, at this stage. Let's simply keep two adjacent directories, e.g. axpy and spatz_axpy.
2. Ideally we should align with the data generation and verification infrastructure used by the other kernels. In the case of different implementations of the same kernel, e.g. axpy and spatz_axpy, to avoid copy-pasting the same scripts around we can symlink them.

colluca (commented):
One more question which just came to my mind, why exactly do we need `riscv_instr.sv` in Snitch to include the vector instructions if we are not decoding them here?

Similarly, why do we need to use Snitch's `riscv_instr.sv` instead of Spatz' own file, within Spatz?

Or is the issue simply that we can't compile two packages with the same name or something like that?

hw/snitch/src/snitch.sv
DiyouS: I believe I turned off the FP_EN so that floating point instruction can be offloaded to Spatz FPU sequencer. But it causes the Caq to be off, triggering some assertion.

hw/snitch/src/snitch.sv
DiyouS: Good catch, this is a leftover part I added for debugging. We can either remove this part or move to tracing

hw/snitch_cluster/src/snitch_cluster.sv
DiyouS: The challenge for double bw is that it changes the interface, especially inside spatz. I can see if it is possible to avoid ifdef at least in snitch_cluster level

make/common.mk
DiyouS: Spatz's [bender file](https://github.com/pulp-platform/spatz/blob/main/Bender.yml) needs this bender flag to include its file. But I can change it in spatz_vpu

make/vsim.mk
DiyouS: I will double check it, but if I remember correctly this comes from some structure type mismatch with amo

sw/runtime/src/start.c
DiyouS: Yep, I will double check

DiyouS:
> One more question which just came to my mind, why exactly do we need `riscv_instr.sv` in Snitch to include the vector instructions if we are not decoding them here?
> 
> Similarly, why do we need to use Snitch's `riscv_instr.sv` instead of Spatz' own file, within Spatz?
> 
> Or is the issue simply that we can't compile two packages with the same name or something like that?

Because currently this file overwrites the one in spatz_vpu, and Spatz's decoder needs to decode these instructions. And the one in Spatz does not contain some features snitch cluster needed, e.g. SSR, post-increment

make/common.mk
colluca: Yeah, that's what I mean, if there's no particular reason for that, it's better to just remove it in there. Especially if all files are gated by that target, it means it's not really necessary.

colluca:
> Because currently this file overwrites the one in spatz_vpu, and Spatz's decoder needs to decode these instructions. And the one in Spatz does not contain some features snitch cluster needed, e.g. SSR, post-increment

What I mean to ask is: what goes wrong when each repo has its own riscv_instr.sv file and no override is done (either through Bender or manually copying the file)? Does Snitch's file end up taking precedence during compilation such that also Spatz picks it up? Is that what you mean by "overwrites"? And if so, how is that a problem if manually overwriting Spatz' file we are achieving the same?

The only condition in which I see issues is if Spatz' overwrites Snitch's during compilation.

DiyouS:
> > Because currently this file overwrites the one in spatz_vpu, and Spatz's decoder needs to decode these instructions. And the one in Spatz does not contain some features snitch cluster needed, e.g. SSR, post-increment
> 
> What I mean to ask is: what goes wrong when each repo has its own riscv_instr.sv file and no override is done (either through Bender or manually copying the file)? Does Snitch's file end up taking precedence during compilation such that also Spatz picks it up? Is that what you mean by "overwrites"? And if so, how is that a problem if manually overwriting Spatz' file we are achieving the same?
> 
> The only condition in which I see issues is if Spatz' overwrites Snitch's during compilation.

I think this is what happens in Gwaihir that Spatz' riscv_instr.sv is compiled later and override Snitch's during compilation

colluca:
> > The only condition in which I see issues is if Spatz' overwrites Snitch's during compilation.
> 
> I think this is what happens in Gwaihir that Spatz' riscv_instr.sv is compiled later and override Snitch's during compilation

Okay, this is what we need to clarify. I believe Bender determines the compilation order according to the dependency DAG, so what you say shouldn't be the case. But I think the way Bender is used in Gwaihir is for it to produce a single compilation command. In that case, probably it can't handle name conflicts (e.g. multiple packages with the same name), and it might be that it filters out all but the first file with the same basename.

If this is the correct explanation, then the best solution would just be to namespace the `riscv_instr.sv` file and package: `snitch_riscv_instr.sv` in Snitch, and `spatz_riscv_instr.sv` in Spatz.

colluca:
> Okay, this is what we need to clarify. I believe Bender determines the compilation order according to the dependency DAG, so what you say shouldn't be the case. But I think the way Bender is used in Gwaihir is for it to produce a single compilation command. In that case, probably it can't handle name conflicts (e.g. multiple packages with the same name), and it might be that it filters out all but the first file with the same basename.
> 
> If this is the correct explanation, then the best solution would just be to namespace the `riscv_instr.sv` file and package: `snitch_riscv_instr.sv` in Snitch, and `spatz_riscv_instr.sv` in Spatz.

I renamed Snitch's file to `snitch_riscv_instr` which should resolve all issues.

PR #322, opened 2026-06-09
idma: Streamline TCDM connection, enable multi-channel operation (NEW)

gbellocchi:
### **Description**
This PR updates the iDMA integration within the Snitch cluster, as shown in Figures 1 and 2.
It is a refreshed version of PR #238 , which had fallen behind the main snitch cluster branch due to being open for a long period.

New features include:
* The wide SoC port directly connects to the TCDM subsystem, allowing external data access while DMA transfers run.
* The wide DMA XBAR has been simplified.
* The DMA uses the OBI protocol to connect to the TCDM subsystem.
* Arbitration between the channels, the superbanks, and the SoC port happens now in TCDM, increasing flexibility and throughput. With minimal changes, it is possible to have multiple DMA cores in the cluster. 
* Add `DMINIT` support in Snitch cluster.
* Add support for OBI events and performance counters.
* Add L1-to-L1 and TLS tests to investigate TCDM deadlock.

### **Tasks**
I have collected the tasks and reviewer comments/suggestions of [PR #88](https://github.com/pulp-platform/iDMA/pull/88):
- [x] Apply modifications of PR #238 to the most recent version of the `snitch_cluster`.
- [x] Merge the corresponding [`idma` branch](https://github.com/pulp-platform/iDMA/tree/luzil/interl) into `devel` ([PR #88](https://github.com/pulp-platform/iDMA/pull/88)).
- [x] Update `custom_instructions.md`.
- [x] Cut `wide_in` by default.
- [x] Fix failing Snitch CI tests.
- [ ] Fix linting issues.
- [ ] Check for (performance) regressions.
- [ ] Check TCDM interconnect area utilization.

<figure>
  snitch_cluster_idma_pr_pre
  <figcaption><b>Figure 1:</b> Old integration of `idma`. The wide AXI4 XBAR interconnect is used for: 
  (i) DMA interfaces to TCDM and SoC;
  (ii) NoC wide in/out; 
  (iii) I$;
  (iv) Zero memory;
  and (v) BootROM.
  </figcaption>
</figure>

<hr>

<figure>
  snitch_cluster_idma_pr_post
  <figcaption><b>Figure 2:</b> New integration of `idma`. The wide AXI4 XBAR interconnect is simplified: 
  (i) DMA interfaces to TCDM via OBI and wide AXI4 requests are transmitted to other clusters via the AXI4 XBAR (NoC wide out);
[...]

PR #330, opened 2026-08-04
idma: Streamline TCDM connection, enable multi-channel operation

gbellocchi:
### **Description**
This PR updates the iDMA integration within the Snitch cluster, as shown in Figures 1 and 2.
It is a refreshed version of PR #238 , which had fallen behind the main snitch cluster branch due to being open for a long period, and a continuation of PR #322.

New features include:
* The wide SoC port directly connects to the TCDM subsystem, allowing external data access while DMA transfers run.
* The wide DMA XBAR has been simplified.
* The DMA uses the OBI protocol to connect to the TCDM subsystem.
* Arbitration between the channels, the superbanks, and the SoC port happens now in TCDM, increasing flexibility and throughput. With minimal changes, it is possible to have multiple DMA cores in the cluster. 
* Add `DMINIT` support in Snitch cluster.
* Add support for OBI events and performance counters.
* Add L1-to-L1 and TLS tests to investigate TCDM deadlock.
* Support iDMA after `common-cells-v2` update.

### **Tasks**
I have collected the tasks and reviewer comments/suggestions of [PR #88](https://github.com/pulp-platform/iDMA/pull/88):
- [x] Apply modifications of PR #238 to the most recent version of the `snitch_cluster`.
- [x] Merge the corresponding [`idma` branch](https://github.com/pulp-platform/iDMA/tree/luzil/interl) into `devel` ([PR #88](https://github.com/pulp-platform/iDMA/pull/88)).
- [x] Update `custom_instructions.md`.
- [x] Cut `wide_in` by default.
- [x] Fix failing Snitch CI tests.
- [x] Test iDMA after `common-cells-v2` update.
- [x] Fix linting issues.
- [x] Fix Spyglass violation in `common_cells`.
- [x] Fix Spyglass violation in `idma`.
- [x] Finalizing [`idma` PR #222](https://github.com/pulp-platform/iDMA/pull/222).
- [ ] Complete PR review.
- [ ] Check for (performance) regressions.
- [ ] Check TCDM interconnect area utilization.

<figure>
  snitch_cluster_idma_pr_pre
  <br><br>
  <figcaption><b>Figure 1:</b> Old integration of `idma`. The wide AXI4 XBAR interconnect is used for: 
  (i) DMA interfaces to TCDM and SoC;
  (ii) NoC wide in/out; 
  (iii) I$;
  (iv) Zero memory;
  and (v) BootROM.
  </figcaption>
</figure>

<hr>

<figure>
  snitch_cluster_idma_pr_post
  <br><br>
  <figcaption><b>Figure 2:</b> New integration of `idma`. The wide AXI4 XBAR interconnect is simplified: 
  (i) DMA interfaces to TCDM via OBI and wide AXI4 requests are transmitted to other clusters via the AXI4 XBAR (NoC wide out);
  (ii) Remove AXI4 TCDM port and directly interface the NoC wide inputs of external DMAs to the TCDM subsystem (bypassing the XBAR);
  (iii) Remove zero memory as the iDMA now supports memset initialization of TCDM.
  </figcaption>
</figure>


hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -37,6 +37,14 @@ module snitch_cc
...
   parameter type         axi_req_t          = logic,
   parameter type         axi_rsp_t          = logic,
+  parameter type         init_req_chan_t    = logic,
+  parameter type         init_rsp_chan_t    = logic,
+  parameter type         init_req_t         = logic,
+  parameter type         init_rsp_t         = logic,
+  parameter type         obi_a_chan_t       = logic,
+  parameter type         obi_r_chan_t       = logic,
+  parameter type         obi_req_t          = logic,
+  parameter type         obi_rsp_t          = logic,
```
colluca, line 43: Can we use the reqrsp interface macros (from the reqrsp_interface repo) to internally derive the req_t and rsp_t types for the OBI and INIT interfaces from the chan_t types? Would simplify the CC parameter list.

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -165,6 +176,8 @@ module snitch_cc
   // Cluster HW barrier
   output logic                              barrier_o,
   input  logic                              barrier_i,
+  // Address decode map
+  input  addr_rule_t [TCDMAliasEnable:0]    dma_addr_rule_i,
```
colluca: Can we name it dma_addr_map_i to reflect the dma's port name?

hw/snitch_cluster/src/snitch_cc.sv
colluca, line 1: This file can be deleted. It was probably restored by accident during the merge with the latest CC changes.

hw/snitch_cluster/src/snitch_cluster_pkg.sv.tpl
colluca, line 1: This file can also be deleted I believe, same as above.

hw/snitch_cluster/src/snitch_cluster_wrapper.sv.tpl
colluca, line 1: Delete, same as above.

make/vsim.mk
```diff
@@ -114,7 +114,9 @@ $(SN_BIN_DIR)/$(TARGET).vsim: $(SN_VSIM_BUILDDIR)/compile.vsim.tcl $(SN_TB_CC_SO
 	@echo "#!/bin/bash" > $@.gui
 	@echo 'binary=$$(realpath $$1)' >> $@.gui
 	@echo 'echo $$binary > .rtlbinary' >> $@.gui
-	@echo '$(SN_VSIM) +permissive $(SN_VSIM_FLAGS) \
+	@echo 'WAVE_ARGS=()' >> $@.gui
+	@echo '[ -n "$$QUESTA_WAVE_DO" ] && WAVE_ARGS=(-do "do $$QUESTA_WAVE_DO")' >> $@.gui
+	@echo '$(SN_VSIM) +permissive "$${WAVE_ARGS[@]}" $(SN_VSIM_FLAGS) \
```
colluca: I assume you used this for debugging, but I think we can remove it

sw/runtime/src/dma.h
```diff
@@ -13,6 +13,22 @@
...
+#define DMCPY_FUNCT7 0b0000011
+#define DMSTATI_FUNCT7 0b0000100
+#define DMSTAT_FUNCT7 0b0000101
+#define DMSTR_FUNCT7 0b0000110
+#define DMREP_FUNCT7 0b0000111
+#define DMINIT_FUNCT7 0b0001001
+
+#define R_TYPE_ENCODE(funct7, rs2, rs1, funct3, rd, opcode)                    \
+    ((funct7 << 25) | (rs2 << 20) | (rs1 << 15) | (funct3 << 12) | (rd << 7) | \
+     (opcode))
```
colluca: Instead of adding all this stuff back, let's add support for DMINIT to the compiler. With the new autogenerated toolchain it should be as trivial as simply bumping riscv-opcodes.

sw/runtime/src/dma.h
```diff
@@ -460,6 +476,64 @@ inline void snrt_dma_memset(void *ptr, uint8_t value, uint32_t len) {
...
 }
 
+/**
+ * @brief Fast memset function performed by DMA with the dminit instruction.
+ * @param ptr Pointer to the start of the region.
+ * @param value Value to set.
+ * @param size The size of the transfer in bytes.
+ * @param channel The index of the channel.
+ */
+inline uint32_t snrt_dma_memset_init_1d(uint64_t ptr, uint8_t value,
```
colluca: What's different from snrt_dma_memset? Can't we just update that function to use DMINIT?

sw/tests/src/deprecated/zero_mem.c
colluca, line 1: No need for deprecating, let's just delete.

colluca (changes requested):
Still need to finish but a first few comments.

sw/runtime/src/dma.h
colluca: In fact I think the compiler we are using already has the DMINIT instruction

hw/snitch_cc/src/snitch_cc.sv
gbellocchi: I think we cannot do this for OBI because its req/rsp are exposed at the interface of `snitch_cc`. We might instead do this for INIT since those are used when `idma` is instantiated in the module body.

gbellocchi:
**Status update**: Latest CI passes, including `spyglass`. I am now looking into:
- Replying to @colluca comments and implementing his suggestions;
- Finalizing [`idma` PR #222](https://github.com/pulp-platform/iDMA/pull/222).

hw/snitch_cc/src/snitch_cc.sv
gbellocchi: Addressed! In the end I moved the INIT definition within the `snitch_cc` since this is only used there.

sw/runtime/src/dma.h
gbellocchi: Yes, I have run a test with release [22.1.7-pulp-0.2.1](https://github.com/pulp-platform/llvm-project/releases/tag/22.1.7-pulp-0.2.1) and it builds and run fine. Actually, it works fine also with `22.1.7-pulp-0.1.0` (targeted by `iis-setup.sh`), so all good!

sw/runtime/src/dma.h
gbellocchi: You're right, there is no need to mantain two different API. I merged the new `dminit`-based implementation into a single `snrt_dma_memset` API and updated kernels and tests.

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -557,26 +591,38 @@ module snitch_cc
...
-      .dma_events_t   (dma_events_t)
+      .AxiAddrWidth     (AddrWidth),
+      .AxiDataWidth     (DMADataWidth),
+      .AxiIdWidth       (DMAIdWidth),
+      .AxiUserWidth     (DMAUserWidth),
+      .NumAxInFlight    (DMANumAxInFlight),
+      .DMAReqFifoDepth  (DMAReqFifoDepth),
+      .NumChannels      (DMANumChannels),
+      .TCDMAliasEnable  (TCDMAliasEnable),
+      .DMATracing       (1),
```
colluca, line 602: What is this needed for?

hw/snitch_cluster/src/snitch_cluster_pkg.sv
```diff
@@ -24,15 +24,17 @@ package snitch_cluster_pkg;
...
+    ToSoC     = 1
+  } dma_e;
+
   typedef enum int unsigned {
-    SoCDMAIn = 32'd0,
-    SDMAMst  = 32'd1,
-    ICache   = 32'd2
+    SDMAMst  = 32'd0,
+    ICache   = 32'd1
   } cluster_master_dma_e;
```
colluca: I think this becomes useless and we can remove it: ICache is never used (and couldn't be, since the number of DMA master is configurable so ICache=1 doesn't represent anything useful) and SDMAMst is just zero now.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -622,6 +641,8 @@ module snitch_cluster
   assign cluster_periph_start_address = IntBootromEnable ? bootrom_end_address : tcdm_end_address;
   assign cluster_periph_end_address   = cluster_periph_start_address + ClusterPeriphSize * 1024;
 
+  // WIP: Modify address space to definitely remove zero memory addresses.
+  localparam int unsigned ZeroMemorySize = 64;
```
colluca: Can we remove this?

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -645,7 +666,6 @@ module snitch_cluster
 
   localparam addr_t ZeroMemAliasStart = PeriphAliasEnd;
   localparam addr_t ZeroMemAliasEnd   = PeriphAliasEnd + ZeroMemorySize * 1024;
```
colluca: Can we remove this and all other references to the zero memory everywhere?

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -747,8 +777,8 @@ module snitch_cluster
     .rst_ni (rst_ni),
     .slv_req_i (wide_in_req_i),
     .slv_resp_o (wide_in_resp_o),
-    .mst_req_o (wide_axi_mst_req[SoCDMAIn]),
-    .mst_resp_i (wide_axi_mst_rsp[SoCDMAIn])
+    .mst_req_o (soc_in_axi_req),
+    .mst_resp_i (soc_in_axi_rsp)
```
colluca: Can we keep the naming aligned with the existing signals? e.g. rename to wide_in_req_q, wide_in_rsp_q?

hw/snitch_cluster/src/snitch_cluster_pkg.sv
```diff
@@ -24,15 +24,17 @@ package snitch_cluster_pkg;
   // Slaves on Cluster DMA AXI Bus
   typedef enum int unsigned {
     SoCDMAOut  = 0,
-    TCDMDMA    = 1,
-    ZeroMemory = 2,
-    Bootrom    = 3
+    BootRom    = 1
```
colluca, line 27: Let's align the naming with all the other parameters: either BootRom everywhere or Bootrom everywhere. I believe the latter is used everywhere else, e.g. BootromEnable, BootromAlias, etc... If in writing we would refer to it as bootrom it should be capitalized as Bootrom. If in writing we would refer to it as boot ROM, it should be capitalized as BootRom. I personally prefer the prior cause I think it's more simple/intuitive, but I don't have a strong opinion, so long as it's consistent. And I believe it used to be consistent so we could also just revert this change.

hw/snitch_cluster/src/snitch_cluster_pkg.sv
```diff
@@ -24,15 +24,17 @@ package snitch_cluster_pkg;
...
   } cluster_slave_dma_e;
 
+    typedef enum logic {
+    TCDMDMA   = 0,
```
colluca, line 31: Capitalize new acronyms according to Google styleguide `TcdmDma`.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -833,38 +875,22 @@ module snitch_cluster
...
+    .axi_req_t (axi_mst_dma_req_t),
+    .axi_resp_t (axi_mst_dma_resp_t),
     .AddrWidth (PhysicalAddrWidth),
     .AxiDataWidth (WideDataWidth),
-    .IdWidth (WideIdWidthOut),
+    .IdWidth (WideIdWidthIn),
     .MemDataWidth (WideDataWidth),
     .BufDepth (MemoryMacroLatency + 1)
-  ) i_axi_to_mem_split (
+  ) i_axi_to_mem_split_soc_in (
```
colluca: Would remove the suffix since there's only one of these modules anyways.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -881,9 +907,58 @@ module snitch_cluster
     assign ext_dma_req[i].q.amo = lsu_pkg::AMONone;
     assign ext_dma_req[i].q.user = '0;
   end
+  localparam int unsigned NumDMAIcoInputs = DMANumChannels + 2;
```
colluca, line 902: See other comment on acronym capitalization.

colluca: Also move this definition to section below (tcdm arbiter) or to top of file.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -881,9 +907,58 @@ module snitch_cluster
...
+        .MemRespLat (MemoryMacroLatency),
+        .NumChannels (DMANumChannels)
+      ) i_obi_to_tcdm (
+        .clk_i,
+        .rst_ni,
+        .obi_req_i (obi_dma_req[i]),
+        .obi_rsp_o (obi_dma_res[i]),
+        .tcdm_req_o (tcdm_dma_req),
+        .tcdm_rsp_i (tcdm_dma_rsp)
+      );
```
colluca, line 937: Can we move this inside the CC and expose directly the TCDM interface? It should then be possible to move all the OBI type definitions and such within the CC (since OBI is only used by the CC anyways), and keep the cluster cleaner.

We would still need the tie-off of the wide TCDM interfaces here.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -881,9 +907,58 @@ module snitch_cluster
...
+    end else begin : gen_dma_obi_to_tcdm_stub
+      // No DMA on this core: tie off its response port.
+      assign obi_dma_res[i] = '0;
+    end
+  end
+  // When no core has DMA, tie off to avoid undriven networks.
+  if (!HasDmaCore) begin : gen_dma_bus_stub
+    assign tcdm_dma_req = '0;
+    assign dma_events   = '0;
```
colluca, line 946: Tie off `dma_events` closer to where they are assigned.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -454,6 +454,14 @@ module snitch_cluster
     return cnt;
   endfunction
 
+  // Reconstruct a per-core bitmask of which cores enable the `Xdma` ISA extension.
+  function automatic bit [NrCores-1:0] xdma_mask();
+    bit [NrCores-1:0] mask = '0;
+    for (int i = 0; i < NrCores; i++)
+      if (IsaCfg[i].Xdma) mask[i] = 1'b1;
+    return mask;
+  endfunction
```
colluca, line 463: The mask is not really used, only for `HasDmaCore`. I would just define a function `supports_xdma` that implements this directly. See `supports_xpulp`.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -576,6 +594,7 @@ module snitch_cluster
       logic r_valid,  r_ready,  r_done, r_bw;
       logic w_valid,  w_ready,  w_done, w_bw;
       logic b_valid,  b_ready,  b_done;
+      logic obi_wr_req, obi_rd_req;
       logic dma_busy;
       axi_pkg::len_t aw_len, ar_len;
       axi_pkg::size_t aw_size, ar_size;
```
colluca, line 600: This whole struct should probably be defined by the DMA, but unless Daniel wants to move this there we can keep it here for now.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -895,12 +970,26 @@ module snitch_cluster
...
+  for (genvar i = 0; i < DMANumChannels; i++) begin : gen_dma_req_tcdm
+    assign dma_interconnect_req[2 + i] = tcdm_dma_req[i];
+  end
+
+  assign ext_dma_rsp[0] = dma_interconnect_rsp[0];
+  assign ext_dma_rsp[1] = dma_interconnect_rsp[1];
+
+  for (genvar i = 0; i < DMANumChannels; i++) begin : gen_dma_rsp_tcdm
+    assign tcdm_dma_rsp[i] = dma_interconnect_rsp[2 + i];
+  end
```
colluca, line 983: This is really a nitpick, but since I read files from top to bottom, I prefer to assign signals before they are used when this is possible. So I would move this block before `i_dma_interconnect`.

hw/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -1134,6 +1223,10 @@ module snitch_cluster
       .axi_aw_chan_t (axi_mst_dma_aw_chan_t),
       .axi_req_t (axi_mst_dma_req_t),
       .axi_rsp_t (axi_mst_dma_resp_t),
+      .obi_a_chan_t (obi_a_chan_t),
+      .obi_r_chan_t (obi_r_chan_t),
+      .obi_req_t (obi_dma_req_t),
+      .obi_rsp_t (obi_dma_rsp_t),
```
colluca, line 1221: If we move the OBI to TCDM module inside the CC this addition should no longer be necessary.

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -118,6 +122,7 @@ module snitch_cc
     SpatzDoubleBw
   ),
   localparam type addr_t = logic [AddrWidth-1:0],
+  parameter type addr_rule_t = axi_pkg::xbar_rule_64_t,
```
colluca, line 125: Move before derived parameter list

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -193,6 +202,31 @@ module snitch_cc
...
+      init_req_chan_t req_chan;
+      logic           req_valid;
+      logic           rsp_ready;
+  } init_req_t;
+
+  typedef struct packed {
+      init_rsp_chan_t rsp_chan;
+      logic           rsp_valid;
+      logic           req_ready;
+  } init_rsp_t;
```
colluca, line 228: Regarding these, I think it would also be better to move them to the DMA repo, if Daniel would be willing to do this. Basically moving them to a typedef.svh file like other repos do. Could be based on the reqrsp_interface repo to make it super simple and uniform with other interfaces. Otherwise this is also fine.

sw/kernels/blas/gemm/src/gemm.h
```diff
@@ -384,15 +384,15 @@ static inline int gemm(const gemm_args_t *args) {
...
-                                lc[c_buff_idx], snrt_cluster()->zeromem.mem,
-                                tile_c_size,
-                                banks_per_buffer * SNRT_TCDM_BANK_WIDTH,
-                                SNRT_TCDM_HYPERBANK_WIDTH);
+                            uint32_t row_size =
+                                banks_per_buffer * SNRT_TCDM_BANK_WIDTH;
+                            uint32_t c_span = (tile_c_size / row_size) *
+                                              SNRT_TCDM_HYPERBANK_WIDTH;
+                            snrt_dma_memset((uint64_t)lc[c_buff_idx], 0, c_span,
+                                            0);
```
colluca, line 392: I don't think this is correct. The previous function was initializing a 2D array in the TCDM (with a particular stride). `snrt_dma_memset` cannot do this. Does the DMA even support this? Otherwise we need to emulate it in software, with a loop of 1D initializations.

sw/kernels/blas/gemm/src/gemm.h
```diff
@@ -384,15 +384,15 @@ static inline int gemm(const gemm_args_t *args) {
...
                         } else {
-                            snrt_dma_start_1d(lc[c_buff_idx],
-                                              snrt_cluster()->zeromem.mem,
-                                              tile_c_size);
+                            snrt_dma_memset((uint64_t)lc[c_buff_idx], 0,
+                                            tile_c_size, 0);
```
colluca, line 395: Since it defaults to zero anyways, we don't need to pass the last argument. Could confuse someone who doesn't remember the function signature by heart.

sw/runtime/src/dma.h
```diff
@@ -434,32 +434,70 @@ inline void snrt_dma_stop_tracking() {
...
+ *          instruction, so no source data is read from memory. The function
+ *          blocks until the region is filled.
+ * @note The function passes the @p channel argument as an immediate,
+ *       thus this must be known at compile time. As a consequence, the
+ *       function must use internal linkage (`static` keyword) and must be
+ *       always inlined. This is true also for all functions invoking this
+ *       function, and passing down an argument to @p channel.
  */
-inline void snrt_dma_memset(void *ptr, uint8_t value, uint32_t len) {
+static inline void snrt_dma_memset(uint64_t ptr, uint8_t value, uint32_t size,
```
colluca, line 448: I think the `static` attribute on the function and the `const` attribute on the channel are legacy. We should align this with the other function signatures, e.g. `snrt_dma_start_1d`.

sw/runtime/src/dma.h
```diff
@@ -434,32 +434,70 @@ inline void snrt_dma_stop_tracking() {
...
+        asm volatile("dminit %[txid], %[size], (%[channel] << 2) | 0 \n"
+                     : [ txid ] "=r"(txid)
+                     : [ size ] "r"(size), [ channel ] "i"(channel));
+    } else if (value == 0xff) {
+        // Initialize with value 1.
+        asm volatile("dminit %[txid], %[size], (%[channel] << 2) | 1 \n"
+                     : [ txid ] "=r"(txid)
+                     : [ size ] "r"(size), [ channel ] "i"(channel));
+    } else {
+        // Initialize with value arbitrary value.
```
colluca: typo "Initialize with arbitrary value"

sw/runtime/src/dma.c
```diff
@@ -19,8 +19,6 @@ extern void snrt_dma_start_tracking();
 
 extern void snrt_dma_stop_tracking();
```
colluca, line 21: Restore once `static` attribute removed.

sw/tests/src/dma_l1l1.c
```diff
@@ -0,0 +1,42 @@
...
+#include <snrt.h>
+
+#define MAXN 1024  // words
+
+int main() {
+#ifdef SNRT_SUPPORTS_DMA
+    if (!snrt_is_dm_core()) return 0;  // only the DMA core
+
+    uint32_t *src = (uint32_t *)snrt_l1_alloc(MAXN * sizeof(uint32_t));
+    uint32_t *dst = (uint32_t *)snrt_l1_alloc(MAXN * sizeof(uint32_t));
```
colluca, line 22: Would be best to use the v2 alloc functions. The v1 functions are deprecated so I can't guarantee they work in all cases.

sw/tests/src/dma_l1l1_remote.c
```diff
@@ -0,0 +1,68 @@
...
+#define MAXN 256
+
+int main() {
+#ifdef SNRT_SUPPORTS_DMA
+    if (!snrt_is_dm_core()) return 0;
+    if (snrt_cluster_num() < 2) return 0;
+    uint32_t errors = 0;
+
+    // Local cluster buffer. Access it via OBI.
+    uint32_t *local = (uint32_t *)snrt_l1_alloc(MAXN * sizeof(uint32_t));
```
colluca, line 18: Same as comment in dma_l1l1.

test/frep_xs.yaml
```diff
@@ -50,15 +50,14 @@ runs:
   # - elf: ../sw/tests/build/openmp_double_buffering.elf
   #   simulators: [vsim, vcs, verilator]
   - elf: ../sw/tests/build/perf_cnt.elf
-    simulators: [vsim, vcs, verilator]
+  #   simulators: [vsim, vcs, verilator]
```
colluca, line 53: Why this change? AFAIK this was there to exclude GVSoC. Not that it really matters, but just curious.

.gitlab-ci.yml
```diff
@@ -209,14 +209,14 @@ snitch-cluster-spatz-vsim:
 # COPIFT experiments
 snitch-cluster-copift-vsim:
   script:
-    - make CFG_OVERRIDE=experiments/copift/cfg.json vsim -j
+    - make DEBUG=ON CFG_OVERRIDE=experiments/copift/cfg.json vsim -j
```
colluca, line 212: Is this addition legacy? Or why is it needed?

.gitlab-ci.yml
```diff
@@ -209,14 +209,14 @@ snitch-cluster-spatz-vsim:
...
     - cd experiments/copift
     - ./experiments.py experiments.yaml --actions sw run perf -j
 
 # Scalar chaining experiments
 snitch-cluster-chaining-vsim:
   script:
-    - make CFG_OVERRIDE=experiments/copift/cfg.json vsim -j
+    - make DEBUG=ON CFG_OVERRIDE=experiments/copift/cfg.json vsim -j
```
colluca, line 219: Same as above.

Bender.yml
```diff
@@ -34,18 +34,24 @@ dependencies:
     rev: common-cells-v2
   idma:
     git: https://github.com/pulp-platform/idma.git
-    rev: __deploy__9785d30__common-cells-v2
+    rev: __deploy__ba01de2__devel
```
colluca, line 37: Is there a release for this now?

Bender.yml
```diff
@@ -34,18 +34,24 @@ dependencies:
...
     rev: colluca/cvxif-demux-bender-ready
   FPnew:
     git: https://github.com/pulp-platform/cvfpu.git
     rev: bump-common-cells-v2
+  obi:
+    git: https://github.com/pulp-platform/obi.git
+    rev: common-cells-v2
+  register_interface:
+    git: https://github.com/pulp-platform/register_interface.git
+    rev: common-cells-v2
```
colluca, line 54: Where is this needed?

util/sim/sim_utils.py
```diff
@@ -118,9 +118,23 @@ def parser(default_simulator='vsim', simulator_choices=['vsim']):
...
 
 
+def apply_wave_file(args):
+    """Export QUESTA_WAVE_DO so the vsim-gui binary picks it up.
+
+    Nothings happen when --wave-file is not passed or the simulator is not vsim-gui.
+    """
+    if getattr(args, 'wave_file', None) is not None:
+        os.environ['QUESTA_WAVE_DO'] = str(Path(args.wave_file).resolve())
+
```
colluca, line 136: Should this be reverted? I don't see the QUESTA_WAVE_DO variable used anywhere.

colluca (changes requested):

sw/deps/riscv-opcodes
colluca, line 1: As far as I can tell, this is reverting to an older version of riscv-opcodes. Is this a mistake?

hw/snitch_cc/src/snitch_cc.sv
gbellocchi: I think @DanielKellerM is a better catch for this question. :)

hw/snitch_cluster/src/snitch_cluster_pkg.sv
gbellocchi: I checked and I 100% agree. Addressed.

hw/snitch_cluster/src/snitch_cluster.sv
gbellocchi: Indeed. I was temporarily keeping this since I wanted to discuss this with you at a certain point.

gbellocchi: Anyway, I have removed the address mapping rules concerning the `zero-memory`. I also kept other rules as they were originally defined. This way, we do not have to propagate any modifications to the SW and we would have 64kB unmapped between `cluster_periph` and `ext_mem`. Would this work for you @colluca ?

hw/snitch_cluster/src/snitch_cluster.sv
gbellocchi: Yes, see above comments for the cluster address mapping.

hw/snitch_cluster/src/snitch_cluster.sv
gbellocchi: Addressed.

hw/snitch_cluster/src/snitch_cluster.sv
colluca: Yeah, I would remove this, and just leave the ext_mem space consecutive to the cluster_periph space. In software the changes should be pretty simple, you just have to adapt the RDL template.

In the future, if anyone needs a custom alignment for the ext mem region, they can extend this, but for now we don't need it, and it shouldn't even break any software in the repo.

.gitlab-ci.yml
gbellocchi: If I remember correctly, I added this target to make the `copift` and `chaining` flows succeed since both were expecting (but not generating) some performance reports. Anyway, I think you told me this was fixed and upstreamed to `main` so we can try removing the tarfet.

Bender.yml
gbellocchi: @DanielKellerM

Bender.yml
gbellocchi: This dependency was resolved through `idma`. However, this was dropped at a certain point, thus breaking some Snitch components. At that time, `snitch_ssr` tests were the first to break. Hence, I expressed the dependency directly in the cluster.

util/sim/sim_utils.py
gbellocchi: This was supported via a mk target that was then removed [here](https://github.com/pulp-platform/snitch_cluster/pull/330#discussion_r3816154838). I generally like loading custom wave files in RTL simulation, hence why I added support for this at a certain point. Anyway, it is no critical feature.

hw/snitch_cluster/src/snitch_cluster_pkg.sv
gbellocchi: I think this blew up at merge-time. I will take care of aligning to `Bootrom`.

hw/snitch_cluster/src/snitch_cluster.sv
gbellocchi: @DanielKellerM

hw/snitch_cc/src/snitch_cc.sv
gbellocchi: @DanielKellerM

sw/tests/src/dma_l1l1.c
gbellocchi: Would the v2 version be `snrt_l1_alloc_cluster_local` then?

test/frep_xs.yaml
gbellocchi: I think this is a mistake. At a certain point I temporarily removed `perf_cnt` tests as they were based on the old DMA interface, thus they were always failing due to the changes we applied. Once restored, I think I forgot to uncomment the `simulators` line. :D

sw/tests/src/dma_l1l1.c
gbellocchi: Btw I noticed other test, runtime, and kernel components still use `snrt_l1_alloc`, but those seem to be there from before. Is this intended?

hw/snitch_cluster/src/snitch_cluster_pkg.sv
gbellocchi: Aligned to `Bootrom`.

hw/snitch_cluster/src/snitch_cluster.sv
gbellocchi: I modified the naming of `NumDMAIcoInputs` to `NumDmaIcoInputs`, but not those at the interface (as `DMANumChannels`) since they would break upper dependencies at integration-time.

sw/deps/riscv-opcodes
gbellocchi: Surely, I don't remember to have modified this.

sw/kernels/blas/gemm/src/gemm.h
gbellocchi: Mm, it might be even though the CI test were passing anyway. Can you check the new version of the code implementation and tell me what you think. Now `memset` goes row-by-row.

sw/tests/src/dma_l1l1.c
gbellocchi: This, and other SW components I introduced, are now updated with `snrt_l1_alloc_cluster_local`, if you want to check. :)

Bender.yml
DanielKellerM: Yes we can keep it, but its not required in the iDMA anymore

Bender.yml
DanielKellerM: Lets target devel for now until the PR is merged
```suggestion
    rev: __deploy__77b16cf__devel
```

hw/snitch_cluster/src/snitch_cluster.sv
DanielKellerM: Moved it to iDMA `IDMA_TYPEDEF_EVENTS_T` in `idma/typedef.svh` pulp-platform/iDMA#229

hw/snitch_cc/src/snitch_cc.sv
DanielKellerM: Same, now in the iDMA [pulp-platform/iDMA#229](https://github.com/pulp-platform/iDMA/pull/229)

hw/snitch_cc/src/snitch_cc.sv
DanielKellerM: 1. `.TCDMAliasEnable` required for the new address decoder to map OBI / TCDM requests. I would leave this for now like this, but its a bit unnecessary since its the point of this PR to decouple TCDM port from the xbar.
2.  `.DMATracing` this was already here, so nothing new. I have a improvement to remove the hardcoded `1` and use the debug flag [#329](https://github.com/pulp-platform/snitch_cluster/pull/329)

Bender.yml
DanielKellerM: Fixed in the branch

hw/snitch_cluster/src/snitch_cluster.sv
DanielKellerM: Fixed in the branch

hw/snitch_cc/src/snitch_cc.sv
DanielKellerM: Fixed in the branch

sw/deps/riscv-opcodes
DanielKellerM: Its just an old commit, strict ancestor SHA. Commited and fixed.

hw/snitch_cc/src/snitch_cc.sv
colluca: Still not clear why 1 is needed. Isn't passing the addrmap sufficient?

hw/snitch_cluster/src/snitch_cluster_pkg.sv
```diff
@@ -10,30 +10,27 @@ package snitch_cluster_pkg;
...
 
-  typedef enum int unsigned {
-    SoCDMAIn = 32'd0,
-    SDMAMst  = 32'd1,
-    ICache   = 32'd2
-  } cluster_master_dma_e;
+    typedef enum logic {
+    TcdmDma   = 0,
+    ToSoc     = 1
+  } dma_e;
```
colluca: I don't remember if we discussed this anymore now, but `ToSoc` seems to not be used anywhere. I don't see a value in having this enumeration; the name doesn't have any meaning either.

We can just hardcode 0 in the DMA addrmap, where TcdmDma is used. If we wanted to do this properly, the iDMA should probably define such a type:
```
typedef enum logic {
    DmaObi = 0,
    DmaAxi = 1
} dma_path_e;
```

sw/kernels/blas/gemm/src/gemm.h
colluca: Looks good to me

sw/runtime/src/dma.h
```diff
@@ -434,32 +434,69 @@ inline void snrt_dma_stop_tracking() {
...
+ * @param size Number of bytes to set.
+ * @param channel The index of the channel.
+ * @details The region is filled by the DMA itself, through the `dminit`
+ *          instruction, so no source data is read from memory. The function
+ *          blocks until the region is filled.
+ * @note The function passes the @p channel argument as an immediate,
+ *       thus this must be known at compile time. As a consequence, the
+ *       function must always be inlined. This is true also for all functions
+ *       invoking this function, and passing down an argument to @p channel.
  */
```
colluca, line 446: This used to be the way channel was handled in most functions, e.g. the dma_start_1d function. But now we are no longer passing channel as an immediate, which is not very robust.

Can you align the new functions you added to this behaviour?

test/frep_xs.yaml
colluca: Sure, let's just revert this :)

.gitlab-ci.yml
colluca: Are we both referring simply to the addition of the `DEBUG=ON` flag? In any case, yes, I believe we can revert this.

Bender.yml
colluca: What do you mean by the `snitch_ssr` tests would break? What exactly would break and how?

util/sim/sim_utils.py
colluca: Oh okay I see. I'm not exactly sure how this is intended to be used though. In general, it might be a valuable addition, but maybe we can move it to a separate PR, where it can be better documented. Otherwise, I would revert it here.

colluca (changes requested):
We're almost there 😁

hw/snitch_cc/src/snitch_cc.sv
DanielKellerM: yes it is. This can be removed indeed

.gitlab-ci.yml
gbellocchi: Yes, the `DEBUG=ON` was needed to generate the reports. Anyway, I'll revert this! :)

Bender.yml
gbellocchi: I am referring to `hw/snitch_ssr/test/fixture_ssr.sv`, which breaks at `vsim` build time because of the missing `reg_test` defined in [`register_interface`](https://github.com/pulp-platform/register_interface/blob/d6e1d4cdaab7870f4faf3f88a1c788eaf5ac129d/src/reg_test.sv#L14).

gbellocchi: To this end, I removed `register_interface` from the dependency list, rebuilt the HW, and this is the resulting error log for reference:

```
# ** Error: hw/snitch_ssr/test/fixture_ssr.sv(216): (vlog-2164) Class or package 'reg_test' not found.
# ** Error: (vlog-13069) hw/snitch_ssr/test/fixture_ssr.sv(216): near "::": syntax error, unexpected ::.
# ** Error: hw/snitch_ssr/test/fixture_ssr.sv(216): (vlog-13205) Syntax error found in the scope following 'reg_test'. Is there a missing '::'?
# ** Error: hw/snitch_ssr/test/fixture_ssr_streamer.sv(284): (vlog-2164) Class or package 'reg_test' not found.
# ** Error: (vlog-13069) hw/snitch_ssr/test/fixture_ssr_streamer.sv(284): near "::": syntax error, unexpected ::.
# ** Error: hw/snitch_ssr/test/fixture_ssr_streamer.sv(284): (vlog-13205) Syntax error found in the scope following 'reg_test'. Is there a missing '::'?
```

util/sim/sim_utils.py
gbellocchi: Ok, I'm going to revert this then.

hw/snitch_cc/src/snitch_cc.sv
gbellocchi: I'll proceed by removing these.

hw/snitch_cluster/src/snitch_cluster_pkg.sv
gbellocchi: About this, I removed `dma_e` and hardcoded the index to 0 in `dma_addr_map`.

gbellocchi: @DanielKellerM do you think it is valuable to define `dma_path_e` in `idma`?

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -598,6 +598,9 @@ module snitch_cc
       .NumAxInFlight    (DmaNumAxInFlight),
       .DMAReqFifoDepth  (DmaReqFifoDepth),
       .NumChannels      (DmaNumChannels),
+      .NumAddrRules     (1 + TcdmAliasEnable),
+      .EnableTcdmObi    (1'b1),
+      .DMATracing       (1),
```
DanielKellerM, line 603: @gbellocchi @colluca had to restore this

Bender.yml
colluca: I see, thanks a lot @gbellocchi, good to keep this then for now.

util/experiments/run.py
```diff
@@ -17,6 +17,7 @@
 
 SIMULATORS = {
     'vsim': Simulator.QuestaSimulator('snitch_cluster.vsim'),
+    'vsim-gui': Simulator.QuestaSimulator('snitch_cluster.vsim.gui'),
```
colluca: This is still to be reverted

colluca (commented):
Thanks a lot for all the hard work @gbellocchi! There's one last thing to revert that was overlooked, otherwise this looks good.

Before merging however, we should test this in Gwaihir, both functionally (that tests don't break, and in fact I think the collectives will break) and in terms of PPA (running a backend to see that it doesn't degrade any metric).

sw/runtime/src/dma.h
colluca: Oh, I understand that the dminit instruction takes an immediate and not a register. Then we have to keep it this way for the time being.

colluca (approved):
LGTM, let's just merge it with @DanielKellerM's other iDMA changes, test it in Gwaihir, and then we're good to go :)

PR #338, opened 2026-09-21
deps: Bump iDMA to 0.7.0 and add the DMOPC compute path

DanielKellerM:
Bumps iDMA to 0.7.0 and adds the on-the-fly compute path (DMOPC) to the `inst64` frontend. Adds a `dma_enable_compute` cluster json config.


DanielKellerM:
Retargeted to `dca` and rebased onto it; both compute tests pass there.

DMOPC transpose, end to end from a Snitch core:

```
[dma_transpose] mode 2 (4 B elems, 16x16 tile): 4x8 -> 8x4, 60 cycles, ok
[dma_transpose] mode 1 (2 B elems, 32x32 tile): 4x8 -> 8x4, 91 cycles, ok
[dma_transpose] PASS (0 errors)
```

The two cases cover both DMOPC operands: the element-size mode rides `rs1`, the tensor dimensions ride `rs2`. The tile is deliberately non-square, so swapping `tensor_m` and `tensor_n` masks an 8x4 region instead of a 4x8 one and fails; elements are position-coded `r*100 + c`; the padding outside the result is poison-checked; and the result must differ from what a plain copy would leave in exactly the off-diagonal elements, so a passthrough scores zero and fails. Reset state of DMOPC is passthrough, which is why that last check is the one that matters.

MX is unaffected and still byte-exact:

```
[dma_mxquant] fp32 ok (0 errors)
[dma_mxquant] fp16 ok (0 errors)
```

`dma_1d` and `dma_2d` unregressed.

Note: `Lint C/C++ Sources` is red from files inherited with the `dca` base, not from this PR. `sw/kernels/blas/axpy/src/axpy.h` and six files under `sw/saris/` already fail clang-format 10 on plain `dca`; this branch touches none of them.

hw/snitch/src/snitch.sv
```diff
@@ -2829,6 +2830,18 @@ module snitch
                 unsupported_inst = 1'b1;
               end
             end
+            // DMOPC latches the on-the-fly compute configuration; rs1 = op, rs2 = params, no rd.
```
DanielKellerM: Remove verbose comment
```suggestion
```

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -33,6 +33,8 @@ module snitch_cc
   parameter int unsigned DMANumAxInFlight   = 0,
   parameter int unsigned DMAReqFifoDepth    = 0,
   parameter int unsigned DMANumChannels     = 0,
+  /// Enable the DMA compute path (DMOPC).
```
DanielKellerM: Remove verbose comment
```suggestion
```

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,79 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
                                   tile_x1_size_in_banks, tile_x0_size_in_banks,
                                   full_x0_size, prec, tile_ld);
 }
+
+// On-the-fly compute (DMOPC); needs `dma_enable_compute` in the cluster config.
+// DMOPC latches the opcode word from rs1 and the op parameters from rs2, and
+// every subsequent DMCPY/DMCPYI carries them until the next DMOPC. Reset state
+// is a plain copy, so no boot DMOPC is needed. See idma_inst64_compute_pkg for
+// the encoding; the frontend asserts on an opcode byte it cannot decode.
```
DanielKellerM: verbose comment block, must be max 1 line

Bender.yml
```diff
@@ -34,12 +34,15 @@ dependencies:
...
+    rev: __deploy__b770669__devel
     pass_targets: ["snitch_cluster"]
   axi:
     git: https://github.com/colluca/axi.git
     rev: multicast
-  common_cells: 2.0.0-beta.2
+  common_cells: 2.0.0-beta.3
+  register_interface:
+    git: https://github.com/pulp-platform/register_interface.git
+    rev: common-cells-v2
```
DanielKellerM: This is okay because it fixes the CI, but maybe better to create a separate PR to fix the CI?

DanielKellerM (commented):
Address this

sw/runtime/src/dma.h
```diff
@@ -6,6 +6,9 @@
 #include <string.h>
 #endif
 
+// DMOPC opcode bytes and operand field positions, generated by iDMA
```
DanielKellerM: verbose comment block
```suggestion
```

sw/runtime/runtime.mk
```diff
@@ -25,6 +25,8 @@ SN_RUNTIME_INCDIRS += $(SN_RUNTIME_DIR)/api/omp
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_DIR)/vendor/riscv-opcodes
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_SRCDIR)
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_HAL_BUILD_DIR)
+# iDMA generates the DMOPC contract header; only deploy tags carry target/sw
```
DanielKellerM: verbose comment block
```suggestion
```

sw/tests/tests.mk
```diff
@@ -35,7 +35,13 @@ SN_RUNTIME_LD_DEPS = $(SN_RUNTIME_MEMORY_LD) $(SN_RUNTIME_BASE_LD) $(SN_RUNTIME_
 # Outputs #
 ###########
 
+# Downstreams set this OFF and name the tests they want, as with SN_BUILD_APPS
+SN_BUILD_TESTS ?= ON
+
+ifeq ($(SN_BUILD_TESTS), ON)
 SN_TESTS       += $(wildcard $(SN_TESTS_SRCDIR)/*.c)
+endif
+
```
DanielKellerM: this changes dont belong in this PR?

test/dma_compute.yaml
DanielKellerM: Doest this follow the repo way of testing? it doesnt seem so. Probably not mergeable

DanielKellerM (commented):
fix

hw/snitch/src/snitch.sv
```diff
@@ -2829,6 +2830,17 @@ module snitch
...
+              if (Xdma) begin
+                acc_req_o.q.addr = DMA_SS;
+                opa_select       = RegRs1;
+                opb_select       = RegRs2;
+                is_acc_inst      = 1'b1;
+                write_rd         = 1'b0;
+              end else begin
+                unsupported_inst = 1'b1;
+              end
+            end
```
colluca: This is exactly the same as the previous case, should be merged

hw/snitch_cc/src/snitch_cc.sv
```diff
@@ -33,6 +33,12 @@ module snitch_cc
   parameter int unsigned DMANumAxInFlight   = 0,
   parameter int unsigned DMAReqFifoDepth    = 0,
   parameter int unsigned DMANumChannels     = 0,
+  parameter bit          DMAEnableCompute   = 1'b0,
+  parameter bit          DMAComputeTranspose           = 1'b1,
+  parameter bit          DMAComputeMxQuant             = 1'b1,
+  parameter bit          DMAComputeMxDequant           = 1'b1,
+  parameter bit          DMAComputeMxFp16              = 1'b1,
+  parameter bit          DMAComputeTransposeFullDuplex = 1'b1,
```
colluca: Since these are not individually used anywhere, but are later packed and passed to the iDMA, I think it would be best if we would just pack them as early as possible (e.g. in `snitch_cluster_wrapper_pkg`). So that, if new parameters are added to these structs, we only have to make the change in one place (e.g. in `snitch_cluster_wrapper_pkg`), and don't have to add the new parameters everywhere throughout the hierarchy.

sw/runtime/runtime.mk
```diff
@@ -25,6 +25,8 @@ SN_RUNTIME_INCDIRS += $(SN_RUNTIME_DIR)/api/omp
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_DIR)/vendor/riscv-opcodes
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_SRCDIR)
 SN_RUNTIME_INCDIRS += $(SN_RUNTIME_HAL_BUILD_DIR)
+SN_IDMA_ROOT := $(shell $(SN_BENDER) path idma)
+SN_RUNTIME_INCDIRS += $(SN_IDMA_ROOT)/target/sw
```
colluca: Just out of curiosity, were some additional software headers recently added to the iDMA? Or is it just the `idma_compute.h` header?

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+
+#ifdef SNRT_SUPPORTS_DMA_COMPUTE
+
+/// On-the-fly compute opcode byte, rs1[7:0]
+#define SNRT_DMA_OPCODE_PASSTHROUGH IDMA_DMOPC_OPC_PASSTHROUGH
+#define SNRT_DMA_OPCODE_MX_QUANT IDMA_DMOPC_OPC_MX_QUANT
+#define SNRT_DMA_OPCODE_MX_DEQUANT IDMA_DMOPC_OPC_MX_DEQUANT
+#define SNRT_DMA_OPCODE_MX_QUANT_FP16 IDMA_DMOPC_OPC_MX_QUANT_FP16
+#define SNRT_DMA_OPCODE_MX_DEQUANT_FP16 IDMA_DMOPC_OPC_MX_DEQUANT_FP16
+#define SNRT_DMA_OPCODE_TRANSPOSE IDMA_DMOPC_OPC_TRANSPOSE
```
colluca: I think we can remove these and just use the original IDMA defines. Probably it would be even better in the long term that the iDMA takes over these dma.h and dma.c files and distributes them itself.

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+ * @param params Op-parameter word for rs2; 0 for ops that take none.
+ * @details Latched at the DMOPC handshake and applied until the next DMOPC.
+ */
+inline void snrt_dma_set_opcode_params(uint32_t opcode, uint32_t params) {
+#ifdef SNRT_SUPPORTS_DMA
+    // dmopc is not in riscv-opcodes rv_xdma yet, so there is no mnemonic
+    asm volatile(".insn r 0x2b, 0x0, 0x0a, x0, %[opcode], %[params] \n"
+                 :
+                 : [ opcode ] "r"(opcode), [ params ] "r"(params)
+                 : "memory");
```
colluca: Now it is right? So let's bump the toolchain and use the instruction mnemonics here.

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+#define SNRT_DMA_OPCODE_PASSTHROUGH IDMA_DMOPC_OPC_PASSTHROUGH
+#define SNRT_DMA_OPCODE_MX_QUANT IDMA_DMOPC_OPC_MX_QUANT
+#define SNRT_DMA_OPCODE_MX_DEQUANT IDMA_DMOPC_OPC_MX_DEQUANT
+#define SNRT_DMA_OPCODE_MX_QUANT_FP16 IDMA_DMOPC_OPC_MX_QUANT_FP16
+#define SNRT_DMA_OPCODE_MX_DEQUANT_FP16 IDMA_DMOPC_OPC_MX_DEQUANT_FP16
+#define SNRT_DMA_OPCODE_TRANSPOSE IDMA_DMOPC_OPC_TRANSPOSE
+
+/// Transpose field widths, mirroring idma_pkg mode and TransposeDimWidth
+#define SNRT_DMA_TP_MODE_MASK 0x3u
+#define SNRT_DMA_TP_DIM_MASK 0xFFFu
```
colluca: Comment doesn't really help explain what these are😁 Are these somehow specific to Snitch's configuration of the iDMA? Otherwise can we move these as well to `idma_compute.h`?

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+
+/**
+ * @brief Set the on-the-fly compute configuration for subsequent transfers.
+ * @param opcode Opcode word for rs1: opcode byte in [7:0], element-size mode
+ *        in [17:16]. Use the SNRT_DMA_OPCODE_* macros.
+ * @param params Op-parameter word for rs2; 0 for ops that take none.
+ * @details Latched at the DMOPC handshake and applied until the next DMOPC.
+ */
+inline void snrt_dma_set_opcode_params(uint32_t opcode, uint32_t params) {
+#ifdef SNRT_SUPPORTS_DMA
```
colluca: Shouldn't it be `SNRT_SUPPORTS_DMA_COMPUTE`?

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+    // dmopc is not in riscv-opcodes rv_xdma yet, so there is no mnemonic
+    asm volatile(".insn r 0x2b, 0x0, 0x0a, x0, %[opcode], %[params] \n"
```
colluca: Let's add documentation for this instruction here `docs/rm/hw/custom_instructions.md` similar to the others

sw/runtime/src/dma.h
```diff
@@ -790,3 +793,103 @@ inline snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+ *      uint32_t, uint32_t, uint32_t)
+ */
+inline uint32_t snrt_dma_start_1d_transpose(volatile void *dst,
+                                            volatile void *src, size_t size,
+                                            uint32_t mode, uint32_t tensor_m,
+                                            uint32_t tensor_n,
+                                            uint32_t channel = 0) {
+    return snrt_dma_start_1d_transpose((uint64_t)dst, (uint64_t)src, size, mode,
+                                       tensor_m, tensor_n, channel);
+}
```
colluca: I would drop `1d` from the function names, I just find it confusing since transposition assumes a 2D data structure.

sw/runtime/src/dma.h
colluca: Actually, this seems to be even already covered by the external #ifdef, or am I missing something?

sw/runtime/src/dma.c
```diff
@@ -90,3 +90,28 @@ extern snrt_dma_txid_t snrt_dma_store_2d_tile_from_banks(
...
+                                            uint32_t tensor_n,
+                                            uint32_t channel);
+
+extern uint32_t snrt_dma_start_1d_transpose(volatile void *dst,
+                                            volatile void *src, size_t size,
+                                            uint32_t mode, uint32_t tensor_m,
+                                            uint32_t tensor_n,
+                                            uint32_t channel);
+
+#endif  // SNRT_SUPPORTS_DMA_COMPUTE
```
colluca: Not sure it's the best thing to have the function declarations conditional, rather their content perhaps, like for the other functions. What do you think?

colluca (changes requested):
I need to finish the review still, but here are some first comments.

sw/tests/src/dma_mxquant.c
```diff
@@ -0,0 +1,106 @@
...
+    snrt_dma_start_1d((volatile void *)out, (volatile void *)mx, MX_BYTES);
+    snrt_dma_wait_all();
+    uint32_t dequant_cycles = snrt_mcycle() - c0;
+
+    uint32_t errors = 0;
+    for (size_t i = 0; i < NUM_ELEMS; i++) {
+        if (out[i] != src[i]) {
+            if (errors < 8)
+                printf("%s mismatch at %u: exp %x got %x\n", name, (unsigned)i,
+                       (unsigned)src[i], (unsigned)out[i]);
```
colluca: `printf` takes significant time to execute in simulation, and is anyways not the best tool for debugging/verification. My suggestion would be to remove it everywhere

cfg/dma_compute.json
colluca: I would avoid creating a new configuration file.

Since the DMA compute features are not mutually exclusive with the other features, we can just enable it in the CI configuration. The previously tested features will continue to be tested, and in addition to those the DMA compute features will also be tested.

Potentially, we could even consider just enabling the DMA compute features in the default cluster configuration, if the PPA impact is not significant, otherwise it might be better to just keep the vanilla cluster as default.

colluca: Thinking about it more, let's just add it to the default configuration.

.gitlab-ci.yml
```diff
@@ -174,6 +174,14 @@ snitch-cluster-mchan-vsim:
     - cd test
     - ../util/experiments/run.py dma_mchan.yaml --simulator vsim -j --run-dir runs/vsim
 
+# Test on-the-fly DMA compute
+snitch-cluster-dma-compute-vsim:
+  script:
+    - make CFG_OVERRIDE=cfg/dma_compute.json sw -j
+    - make vsim TRACE=OFF -j
+    - cd test
+    - ../util/experiments/run.py dma_compute.yaml --simulator vsim -j --run-dir runs/vsim
```
colluca: See comment to dma_compute.json

Bender.yml
```diff
@@ -34,12 +34,15 @@ dependencies:
     rev: common-cells-v2
   idma:
     git: https://github.com/pulp-platform/idma.git
-    rev: __deploy__9785d30__common-cells-v2
+    rev: __deploy__b770669__devel
```
colluca: Is there a DMA release with these features now?

colluca (changes requested):

hw/snitch_cc/src/snitch_cc.sv
DanielKellerM: Agree

PR #342, opened 2026-09-23
sw: Reformat the axpy fallback call with clang-format

DanielKellerM:
`Lint C/C++ Sources` is red on every PR against `dca` because `sw/kernels/blas/axpy/src/axpy.h` is not clang-format clean. This is the only offending file; `sw/saris` is covered by `.clang-format-ignore`.


Lore0599 (approved):
LGTM :)

